PR #5, opened 2020-08-10
Fix bug where query is not created correctly when there only hidden args exists

### LINKED ISSUE #3: [BUG] Query is not created properly if there is hidden Args
Bug I found which I mentioned in [slack](https://prestosql.slack.com/archives/CFLB9AMBN/p1596587181197000?thread_ts=1596437047.120200&cid=CFLB9AMBN) at this part of the code
- https://github.com/prestosql/presto-go-client/blob/master/presto/presto.go#L559-L578

### Error when there are only hidden args

If there are no regular args and only hidden Args, query is created like `EXECUTE _presto_go USING`. This causes an error `io.prestosql.sql.parser.ParsingException: line 1:26: mismatched input '<EOF>'. Expecting: <expression>` because it ends with USING.

### Suggest How-to-fix

~~First check if there are any ‘simple’ arguments, then query should be `EXECUTE preparedStatementName USING ….` . 
If there are only hidden args, query should be from original, st.query.~~
As  electrum saids in slack, simply checking length of `ss []string` seems to be simpler and cleaner method

Chaho12:
Issue related to #3 

Query is not created correctly if the arg provided to QueryContext function is hidden arg (PrestoUserHeader) 

Fixes #3

### DIFF
--- a/presto/presto.go
+++ b/presto/presto.go
@@ -574,7 +574,9 @@ func (st *driverStmt) QueryContext(ctx context.Context, args []driver.NamedValue
 				ss = append(ss, s)
 			}
 		}
-		query = "EXECUTE " + preparedStatementName + " USING " + strings.Join(ss, ", ")
+		if len(ss) > 0 {
+			query = "EXECUTE " + preparedStatementName + " USING " + strings.Join(ss, ", ")
+		}
 	}
 
 	req, err := st.conn.newRequest("POST", st.conn.baseURL+"/v1/statement", strings.NewReader(query), hs)
--- a/presto/presto_test.go
+++ b/presto/presto_test.go
@@ -203,6 +203,41 @@ func TestAuthFailure(t *testing.T) {
 	defer db.Close()
 }
 
+func TestQueryWithJustUserHeader(t *testing.T) {
+	c := &Config{
+		PrestoURI:         "http://foobar@localhost:8080",
+		SessionProperties: map[string]string{"query_priority": "1"},
+	}
+	dsn, err := c.FormatDSN()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	db, err := sql.Open("presto", dsn)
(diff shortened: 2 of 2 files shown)


Chaho12:
I sent the signed CLA to cla@prestosql.io 1 hour ago (I guess this is why there is fail report)

losipiuk:
@cla-bot check

Chaho12:
@findepi 
I am not sure if this is a good test sample, but I do have a test code that queries `SELECT 2` with user header.
Want me to add this test code to PR?
``` go
func TestUser(t *testing.T) {
	c := &Config{
		PrestoURI:         "http://foobar@localhost:8080",
		SessionProperties: map[string]string{"query_priority": "1"},
	}
	dsn, err := c.FormatDSN()
	if err != nil {
		t.Fatal(err)
	}

	db, err := sql.Open("presto", dsn)
	if err != nil {
		t.Fatal(err)
	}
	defer db.Close()
	rows, err := db.Query("SELECT 2", sql.Named(PrestoUserHeader, string("TestUser")))
	if err != nil {
		t.Fatal("Failed executing query", err.Error())
	}

	if rows != nil {
		for rows.Next() {
			var ts string
			err = rows.Scan(&ts)
			if err != nil {
				t.Fatal("Failed scanning query result", err.Error())
			}
			want := "2"
			if ts != want {
				t.Fatal("Expected value does not equal result value : ", ts, " != ", want)
			}
		}
	}
}
```

losipiuk:
Please do add the test method. I think you can name it `TestQueryWithJustUserHeader`

losipiuk (approved):

Chaho12:
@losipiuk added test code :) as well as testing it

image

losipiuk:
Merged, thanks!

findepi:
@th0114nd  does this solve your needs?

PR #7, opened 2020-08-11
Fix serializing user name

### LINKED ISSUE #6: Presto User Header is unnecessarily serialized
Value part for `hidden args` is serialized. Here, I think it should be `st.user = arg.Value.(string)` cause `s` value is serialized value and user name should not be serialized (else input name would be  `'Amy'`  and not `Amy` like image below ).
- https://github.com/prestosql/presto-go-client/blob/master/presto/presto.go#L568

### Test

``` go
_, err = db.Query("SELECT 2", sql.Named(PrestoUserHeader, string("Amy")))
```

#### Before : Serialized
image

#### After : Not serialized
image

Chaho12:
Issue related to https://github.com/prestosql/presto-go-client/issues/6 where hidden args should not be serialized. 

### before
image

### after

image

Fixes #6

### DIFF
--- a/presto/presto_test.go
+++ b/presto/presto_test.go
@@ -203,7 +203,7 @@ func TestAuthFailure(t *testing.T) {
 	defer db.Close()
 }
 
-func TestQueryWithJustUserHeader(t *testing.T) {
+func TestQueryForUsername(t *testing.T) {
 	c := &Config{
 		PrestoURI:         "http://foobar@localhost:8080",
 		SessionProperties: map[string]string{"query_priority": "1"},
@@ -218,7 +218,7 @@ func TestQueryWithJustUserHeader(t *testing.T) {
 		t.Fatal(err)
 	}
 	defer db.Close()
-	rows, err := db.Query("SELECT 2", sql.Named("X-Presto-User", string("TestUser")))
+	rows, err := db.Query("SELECT current_user", sql.Named("X-Presto-User", string("TestUser")))
 	if err != nil {
 		t.Fatal("Failed executing query", err.Error())
 	}
@@ -230,7 +230,7 @@ func TestQueryWithJustUserHeader(t *testing.T) {
 			if err != nil {
 				t.Fatal("Failed scanning query result", err.Error())
 			}
-			want := "2"
+			want := "TestUser"
 			if ts != want {
 				t.Fatal("Expected value does not equal result value : ", ts, " != ", want)
 			}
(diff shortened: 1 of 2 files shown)


losipiuk:
@cla-bot check

Chaho12:
@ebyhr 
Unfortunately, I can't add a test case because there is no callback mechanism or sth similar in the code as of now.
In order to test, i need to run the query like `rows, err := db.Query("SELECT 2", sql.Named(PrestoUserHeader, string("TestSample")))` but `rows` only has column fields/types and data.

Chaho12:
@ebyhr Thx for message in slack.

As you said, I've got test case ready, but this test code fails executing query until another BUG is fixed at the following PR
- https://github.com/prestosql/presto-go-client/pull/5 
- works well in my local code where I fixed the bug

``` go
func TestUser(t *testing.T) {
	c := &Config{
		PrestoURI:         "http://foobar@localhost:8080",
		SessionProperties: map[string]string{"query_priority": "1"},
	}
	dsn, err := c.FormatDSN()
	if err != nil {
		t.Fatal(err)
	}

	db, err := sql.Open("presto", dsn)
	if err != nil {
		t.Fatal(err)
	}
	defer db.Close()
	rows, err := db.Query("SELECT current_user", sql.Named("X-Presto-User", string("TestUser")))
	if err != nil {
		t.Fatal("Failed executing query", err.Error())
	}

	if rows != nil {
		for rows.Next() {
			var ts string
			err = rows.Scan(&ts)
			if err != nil {
				t.Fatal("Failed scanning query result", err.Error())
			}
			want := "TestUser"
			if ts != want {
				t.Fatal("Expected value does not equal result value : ", ts, " != ", want)
			}
		}
	}
}
```

### before

image

### after

image

Chaho12:
@ebyhr added test code :)

presto/presto_test.go
```diff
@@ -238,6 +238,41 @@ func TestQueryWithJustUserHeader(t *testing.T) {
 	}
 }
 
+func TestQueryForUsername(t *testing.T) {
```
losipiuk, line 241: I guess you can just change the `TestQueryWithJustUserHeader` to query for `current_user` and have and rename it to `TestQueryForUsername`; but having two tests is fine too.
Please squash two commits together.

Chaho12: Two tests seems to overlap each other, so I got rid of `TestQueryWithJustUserHeader` because `TestQueryForUsername` tests with just user header.

losipiuk (approved):

losipiuk:
/lgtm
/approve
Thanks!

Issue #12, opened 2020-08-26
Return extra runtime statistics information immediately


Chaho12:
Unlike other other drivers which simply uses http request and gets runtime statistics information using HTTP protocol, Presto-Go-Client uses https://golang.org/pkg/database/sql/driver to implement and returns rows only after the query completes

@losipiuk suggests using named query arguments to pass callback to implementation of QueryContext method.
- https://github.com/prestosql/presto-go-client/commit/be6bb15edd764217e6626ceb14f024bdf33cc4d9

### FYI

After setting connection by sql.Open("presto", dsn) whole process of querying data is as follows. These functions are all within go SQL.

`sql.Query -> QueryContext -> query -> queryDC -> rowsiFromStatement -> ctxDriverStmtQuery ->  siCtx.QueryContext(ctx, nvdargs) ->` then returns rows
- queryDC : queryDC executes a query on the given connection.
- siCtx.QueryContext(ctx, nvdargs) : presto-go-client QueryContext function

Relevant slack discussion: https://prestosql.slack.com/archives/CFLB9AMBN/p1597988801033400

losipiuk:
Thanks for the issue. 

I would suggest to model the object representing statistics after what we have for JDBC https://github.com/prestosql/presto/blob/c74b775e103a7a239aafef1b7ca15fe6370495fe/presto-jdbc/src/main/java/io/prestosql/jdbc/QueryStats.java.
And the callback interface could be:

```go
type QueryStatsCallback interface{
   // called on query stats update. Final call will be done just after query execution finished.
   StatsUpdated(QueryStats)

   // value in milliseconds determining how often the stats callback should be called. If 0 it will be called as frequently as we get new stats from the server.
   GetCallBackPeriod() int64
}
```

Then we can register the stats via extra argument passed to `QueryContext`:
```go
db.Query("SELECT 2", sql.Named("X-Query-Stats-Callback", callback))
```

Alternatively we can drop `GetCallBackPeriod` from the interface and use extra `X-Query-Stats-Callback-Period` parameter instead.

@Chaho12 @ebyhr WDYT?

Chaho12:
@losipiuk can you checkout this code? I added `ProgressPercentage` and added your callback code inside `fetch` method. 
- [fetch](https://github.com/prestosql/presto-go-client/blob/master/presto/presto.go#L608) is called inside QueryContext and this is where the actual HTTP Get is called. 

[...]

PR #14, opened 2020-11-09
add callback QueryStats with callback period

Chaho12:
### oss

- #12 

### test

test with 1 second callback period
- presto_test code is set as 0.1 * 1000 or 0.1 seconds for build purpose

Because callback period is set only at `running state` and test query is `SELECT 2`, `RUNNING -> FINISHED` state changes almost immediately so sometimes it passes `RUNNING state` before it fetches the state.
```
=== RUN   TestQueryWithCallbackAsParam
QUEUED, 0 
QUEUED, 0 
QUEUED, 0 
RUNNING, 0 
FINISHED, 100 
--- PASS: TestQueryWithCallbackAsParam (1.09s)
PASS


=== RUN   TestQueryWithCallbackAsParam
QUEUED, 0 
QUEUED, 0 
QUEUED, 0 
FINISHED, 0 
FINISHED, 0 
FINISHED, 100 
--- PASS: TestQueryWithCallbackAsParam (0.12s)
PASS
```

The reason why only `running state` has callback period (am sure you guys know much more about the whole API process) is because presto does not start the job until query is submitted. Here below is screenshot when there was 1 second callback period (query finished in 50ms while test RUN took 1 seconds), which means that this callback period does not delay presto processing time.

스크린샷 2020-11-06 오후 3 11 56

______

#### edit of 2020-11-11

- changed callback period to argument passed instead of forcing user to define function for callback interface
- https://github.com/prestosql/presto-go-client/pull/14/commits/9d1a64d048a6d433138c7a69139d688bf7d68a45

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,6 +8,14 @@ In order to accept your pull request, we need you to [submit a CLA](https://gith
 
 By contributing to Trino, you agree that your contributions will be licensed under the [Apache License Version 2.0 (APLv2)](LICENSE).
 
+# Go Test
+
+Please Run [go test](https://pkg.go.dev/testing) before creating Pull Request
+
+```bash
+go test -v -race -timeout 1m ./...
+```
+
 # Releases
 
 To create a new release, a maintainer with repository write permissions needs to create and push a new git tag.
--- a/trino/trino.go
+++ b/trino/trino.go
@@ -100,6 +100,9 @@ var (
 
 	// ErrInvalidResponseType indicates that the server returned an invalid type definition.
 	ErrInvalidResponseType = errors.New("trino: server response contains an invalid type")
+
+	// ErrInvalidProgressCallbackHeader indicates that server did not get valid headers for progress callback
+	ErrInvalidProgressCallbackHeader = errors.New("trino: both " + trinoProgressCallbackParam + " and " + trinoProgressCallbackPeriodParam + " must be set when using progress callback")
 )
 
 const (
@@ -120,6 +123,9 @@ const (
 	trinoClearSessionHeader    = trinoHeaderPrefix + `Clear-Session`
 	trinoSetRoleHeader         = trinoHeaderPrefix + `Set-Role`
 	trinoExtraCredentialHeader = trinoHeaderPrefix + `Extra-Credential`
+
+	trinoProgressCallbackParam       = trinoHeaderPrefix + `Progress-Callback`
+	trinoProgressCallbackPeriodParam = trinoHeaderPrefix + `Progress-Callback-Period`
 
 	trinoAddedPrepareHeader       = trinoHeaderPrefix + `Added-Prepare`
 	trinoDeallocatedPrepareHeader = trinoHeaderPrefix + `Deallocated-Prepare`
@@ -232,12 +238,14 @@ func (c *Config) FormatDSN() (string, error) {
 
 // Conn is a Trino connection.
 type Conn struct {
-	baseURL         string
-	auth            *url.Userinfo
-	httpClient      http.Client
-	httpHeaders     http.Header
-	kerberosClient  client.Client
-	kerberosEnabled bool
+	baseURL               string
+	auth                  *url.Userinfo
+	httpClient            http.Client
+	httpHeaders           http.Header
+	kerberosClient        client.Client
+	kerberosEnabled       bool
+	progressUpdater       ProgressUpdater
+	progressUpdaterPeriod queryProgressCallbackPeriod
 }
 
 var (
@@ -541,9 +549,11 @@ func newErrQueryFailedFromResponse(resp *http.Response) *ErrQueryFailed {
 }
 
 type driverStmt struct {
-	conn  *Conn
-	query string
-	user  string
+	conn    *Conn
+	query   string
+	user    string
+	statsCh chan QueryProgressInfo
+	doneCh  chan struct{}
 }
 
 var (
@@ -553,7 +563,14 @@ var (
 	_ driver.NamedValueChecker = &driverStmt{}
 )
 
+// Close closes statement just before releasing connection
 func (st *driverStmt) Close() error {
+	if st.doneCh != nil {
+		close(st.doneCh)
+	}
+	if st.statsCh != nil {
+		<-st.statsCh
+	}
 	return nil
 }
 
@@ -576,11 +593,14 @@ func (st *driverStmt) ExecContext(ctx context.Context, args []driver.NamedValue)
 		queryID:      sr.ID,
 		nextURI:      sr.NextURI,
 		rowsAffected: sr.UpdateCount,
+		statsCh:      st.statsCh,
+		doneCh:       st.doneCh,
 	}
 	// consume all results, if there are any
 	for err == nil {
 		err = rows.fetch(true)
 	}
+
 	if err != nil && err != io.EOF {
 		return nil, err
 	}
@@ -593,6 +613,13 @@ func (st *driverStmt) CheckNamedValue(arg *driver.NamedValue) error {
 		return nil
 	}
 	if reflect.TypeOf(arg.Value).Kind() == reflect.Slice {
+		return nil
+	}
+
+	if arg.Name == trinoProgressCallbackParam {
+		return nil
+	}
+	if arg.Name == trinoProgressCallbackPeriodParam {
 		return nil
 	}
 	return driver.ErrSkip
@@ -609,19 +636,20 @@ type stmtResponse struct {
 }
 
 type stmtStats struct {
-	State           string    `json:"state"`
-	Scheduled       bool      `json:"scheduled"`
-	Nodes           int       `json:"nodes"`
-	TotalSplits     int       `json:"totalSplits"`
-	QueuesSplits    int       `json:"queuedSplits"`
-	RunningSplits   int       `json:"runningSplits"`
-	CompletedSplits int       `json:"completedSplits"`
-	UserTimeMillis  int       `json:"userTimeMillis"`
-	CPUTimeMillis   int       `json:"cpuTimeMillis"`
-	WallTimeMillis  int       `json:"wallTimeMillis"`
-	ProcessedRows   int       `json:"processedRows"`
-	ProcessedBytes  int       `json:"processedBytes"`
-	RootStage       stmtStage `json:"rootStage"`
+	State              string    `json:"state"`
+	Scheduled          bool      `json:"scheduled"`
+	Nodes              int       `json:"nodes"`
+	TotalSplits        int       `json:"totalSplits"`
+	QueuesSplits       int       `json:"queuedSplits"`
+	RunningSplits      int       `json:"runningSplits"`
+	CompletedSplits    int       `json:"completedSplits"`
+	UserTimeMillis     int       `json:"userTimeMillis"`
+	CPUTimeMillis      int       `json:"cpuTimeMillis"`
+	WallTimeMillis     int       `json:"wallTimeMillis"`
+	ProcessedRows      int       `json:"processedRows"`
+	ProcessedBytes     int       `json:"processedBytes"`
+	RootStage          stmtStage `json:"rootStage"`
+	ProgressPercentage float32   `json:"progressPercentage"`
 }
 
 type stmtError struct {
@@ -678,6 +706,8 @@ func (st *driverStmt) QueryContext(ctx context.Context, args []driver.NamedValue
 		stmt:    st,
 		queryID: sr.ID,
 		nextURI: sr.NextURI,
+		statsCh: st.statsCh,
+		doneCh:  st.doneCh,
 	}
 	if err = rows.fetch(false); err != nil {
 		return nil, err
@@ -693,6 +723,15 @@ func (st *driverStmt) exec(ctx context.Context, args []driver.NamedValue) (*stmt
 		hs = make(http.Header)
 		var ss []string
 		for _, arg := range args {
+			if arg.Name == trinoProgressCallbackParam {
+				st.conn.progressUpdater = arg.Value.(ProgressUpdater)
+				continue
+			}
+			if arg.Name == trinoProgressCallbackPeriodParam {
+				st.conn.progressUpdaterPeriod.Period = arg.Value.(time.Duration)
+				continue
+			}
+
 			s, err := Serial(arg.Value)
 			if err != nil {
 				return nil, err
@@ -715,6 +754,9 @@ func (st *driverStmt) exec(ctx context.Context, args []driver.NamedValue) (*stmt
 				}
 				ss = append(ss, s)
 			}
+		}
+		if (st.conn.progressUpdater != nil && st.conn.progressUpdaterPeriod.Period == 0) || (st.conn.progressUpdater == nil && st.conn.progressUpdaterPeriod.Period > 0) {
+			return nil, ErrInvalidProgressCallbackHeader
 		}
 		if len(ss) > 0 {
 			query = "EXECUTE " + preparedStatementName + " USING " + strings.Join(ss, ", ")
@@ -739,6 +781,38 @@ func (st *driverStmt) exec(ctx context.Context, args []driver.NamedValue) (*stmt
 	if err != nil {
 		return nil, fmt.Errorf("trino: %v", err)
 	}
+
+	if st.conn.progressUpdater != nil {
+		st.statsCh = make(chan QueryProgressInfo)
+		st.doneCh = make(chan struct{})
+
+		// progress updater go func
+		go func() {
+			for {
+				select {
+				case stats := <-st.statsCh:
+					st.conn.progressUpdater.Update(stats)
+				case <-st.doneCh:
+					close(st.statsCh)
+					return
+				}
+			}
+		}()
+
+		// initial progress callback call
+		srStats := QueryProgressInfo{
+			QueryId:    sr.ID,
+			QueryStats: sr.Stats,
+		}
+		select {
+		case st.statsCh <- srStats:
+		default:
+			// ignore when can't send stats
+		}
+		st.conn.progressUpdaterPeriod.LastCallbackTime = time.Now()
+		st.conn.progressUpdaterPeriod.LastQueryState = sr.Stats.State
+	}
+
 	return &sr, handleResponseError(resp.StatusCode, sr.Error)
 }
 
@@ -754,6 +828,9 @@ type driverRows struct {
 	coltype      []*typeConverter
 	data         []queryData
 	rowsAffected int64
+
+	statsCh chan QueryProgressInfo
+	doneCh  chan struct{}
 }
 
 var _ driver.Rows = &driverRows{}
@@ -976,7 +1053,9 @@ func (qr *driverRows) fetch(allowEOF bool) error {
 	qr.rowindex = 0
 	qr.data = qresp.Data
 	qr.nextURI = qresp.NextURI
+
 	if len(qr.data) == 0 {
+		qr.scheduleProgressUpdate(qresp.ID, qresp.Stats)
 		if qr.nextURI != "" {
 			return qr.fetch(allowEOF)
 		}
@@ -998,6 +1077,7 @@ func (qr *driverRows) fetch(allowEOF bool) error {
 		}
 	}
 	qr.rowsAffected = qresp.UpdateCount
+	qr.scheduleProgressUpdate(qresp.ID, qresp.Stats)
 	return nil
 }
 
@@ -1041,6 +1121,33 @@ func (qr *driverRows) initColumns(qresp *queryResponse) error {
 		}
 	}
 	return nil
+}
+
+func (qr *driverRows) scheduleProgressUpdate(id string, stats stmtStats) {
+	if qr.stmt.conn.progressUpdater == nil {
+		return
+	}
+
+	qrStats := QueryProgressInfo{
+		QueryId:    id,
+		QueryStats: stats,
+	}
+	currentTime := time.Now()
+	diff := currentTime.Sub(qr.stmt.conn.progressUpdaterPeriod.LastCallbackTime)
+	period := qr.stmt.conn.progressUpdaterPeriod.Period
+
+	// Check if period has not passed yet AND if query state did not change
+	if diff < period && qr.stmt.conn.progressUpdaterPeriod.LastQueryState == qrStats.QueryStats.State {
+		return
+	}
+
+	select {
+	case qr.statsCh <- qrStats:
+	default:
+		// ignore when can't send stats
+	}
+	qr.stmt.conn.progressUpdaterPeriod.LastCallbackTime = currentTime
+	qr.stmt.conn.progressUpdaterPeriod.LastQueryState = qrStats.QueryStats.State
 }
 
 type typeConverter struct {
@@ -1942,4 +2049,20 @@ func (s *NullSlice3Map) Scan(value interface{}) error {
 	s.Slice3Map = slice
 	s.Valid = true
 	return nil
+}
+
+type QueryProgressInfo struct {
+	QueryId    string
+	QueryStats stmtStats
+}
+
+type queryProgressCallbackPeriod struct {
+	Period           time.Duration
+	LastCallbackTime time.Time
+	LastQueryState   string
+}
+
+type ProgressUpdater interface {
+	// Update the query progress, immediately when the query starts, when receiving data, and once when the query is finished.
+	Update(QueryProgressInfo)
 }
--- a/trino/trino_test.go
+++ b/trino/trino_test.go
@@ -23,6 +23,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"reflect"
+	"sort"
 	"testing"
 	"time"
 
@@ -246,6 +247,101 @@ func TestQueryForUsername(t *testing.T) {
 		require.NoError(t, rows.Scan(&user), "Failed scanning query result")
 
 		assert.Equal(t, "TestUser", user, "Expected value does not equal result value")
+	}
+}
+
+type TestQueryProgressCallback struct {
+	statusMap map[time.Time]string
+}
+
+func (qpc *TestQueryProgressCallback) Update(qpi QueryProgressInfo) {
+	qpc.statusMap[time.Now()] = qpi.QueryStats.State
+}
+
+func TestQueryProgressWithCallback(t *testing.T) {
+	if testing.Short() {
+		t.Skip("Skipping test in short mode.")
+	}
+	c := &Config{
+		ServerURI:         *integrationServerFlag,
+		SessionProperties: map[string]string{"query_priority": "1"},
+	}
+
+	dsn, err := c.FormatDSN()
+	require.NoError(t, err)
+
+	db, err := sql.Open("trino", dsn)
+	require.NoError(t, err)
+
+	t.Cleanup(func() {
+		assert.NoError(t, db.Close())
+	})
+
+	callback := &TestQueryProgressCallback{}
+
+	_, err = db.Query("SELECT 2", sql.Named("X-Trino-Progress-Callback", callback))
+	assert.EqualError(t, err, ErrInvalidProgressCallbackHeader.Error(), "unexpected error")
+}
+
+func TestQueryProgressWithCallbackPeriod(t *testing.T) {
+	if testing.Short() {
+		t.Skip("Skipping test in short mode.")
+	}
+	c := &Config{
+		ServerURI:         *integrationServerFlag,
+		SessionProperties: map[string]string{"query_priority": "1"},
+	}
+
+	dsn, err := c.FormatDSN()
+	require.NoError(t, err)
+
+	db, err := sql.Open("trino", dsn)
+	require.NoError(t, err)
+
+	t.Cleanup(func() {
+		assert.NoError(t, db.Close())
+	})
+
+	statusMap := make(map[time.Time]string)
+	progressUpdater := &TestQueryProgressCallback{
+		statusMap: statusMap,
+	}
+	progressUpdaterPeriod, err := time.ParseDuration("1ms")
+
+	rows, err := db.Query("SELECT 2",
+		sql.Named("X-Trino-Progress-Callback", progressUpdater),
+		sql.Named("X-Trino-Progress-Callback-Period", progressUpdaterPeriod),
+	)
+	require.NoError(t, err, "Failed executing query")
+	assert.NotNil(t, rows)
+
+	for rows.Next() {
+		var ts string
+		require.NoError(t, rows.Scan(&ts), "Failed scanning query result")
+
+		assert.Equal(t, "2", ts, "Expected value does not equal result value")
+	}
+
+	if err = rows.Err(); err != nil {
+		t.Fatal(err)
+	}
+	if err = rows.Close(); err != nil {
+		t.Fatal(err)
+	}
+
+	// sort time in order to calculate interval
+	var keys []time.Time
+	for k := range statusMap {
+		keys = append(keys, k)
+	}
+	sort.Slice(keys, func(i, j int) bool {
+		return keys[i].Before(keys[j])
+	})
+
+	for i, k := range keys {
+		if i > 0 {
+			assert.GreaterOrEqual(t, k.Sub(keys[i-1]), progressUpdaterPeriod)
+		}
 	}
 }
 


Chaho12:
@losipiuk Hi. I was wondering what further procedures I need to take to apply this pr.

losipiuk:
> @losipiuk Hi. I was wondering what further procedures I need to take to apply this pr.
> I forked this repo and applied this PR to our repo, and using it in our service

Thanks @Chaho12. I did not notice the changes you pushed in February, sorry for that. I will take a look today or tomorrow.

trino/trino.go
```diff
@@ -1649,3 +1725,21 @@ func (s *NullSlice3Map) Scan(value interface{}) error {
 	s.Valid = true
 	return nil
 }
+
+// QueryResponse object used for callback
+type QueryResponse struct {
```
losipiuk, line 1730: Rename to QueryProgressCallbackResponse

trino/trino.go
```diff
@@ -1649,3 +1725,21 @@ func (s *NullSlice3Map) Scan(value interface{}) error {
...
+	Error      stmtError `json:"error"`
+}
+
+type QueryCallbackPeriod struct {
+	Period    float64   `json:"period"`
+	CheckTime time.Time `json:"checkTime"`
+}
+
+// QueryCallback interface for callback
+type QueryCallback interface {
```
losipiuk, line 1742: Rename to QueryProgressCallback

trino/trino.go
```diff
@@ -1649,3 +1725,21 @@ func (s *NullSlice3Map) Scan(value interface{}) error {
...
+	// Called on query stats update, before and after query execution finished.
+	StatsUpdated(QueryResponse)
```
losipiuk, line 1744: Rename to `ProgressUpdated`

trino/trino.go
```diff
@@ -710,6 +762,7 @@ func (qr *driverRows) Close() error {
 	defer cancel()
 	resp, err := qr.stmt.conn.roundTrip(ctx, req)
 	if err != nil {
+		qr.stmt.conn.ResetSession(qr.ctx)
```
losipiuk, line 765: we should not need to call `ResetSession` explicitly.

trino/trino.go
```diff
@@ -1649,3 +1725,21 @@ func (s *NullSlice3Map) Scan(value interface{}) error {
 	s.Valid = true
 	return nil
 }
+
+// QueryResponse object used for callback
+type QueryResponse struct {
+	ID         string    `json:"id"`
+	QueryStats stmtStats `json:"stmtStats"`
```
losipiuk, line 1732: how do you make use of `QueryStats` and `Error` if those are of non-public types?

trino/trino.go
```diff
@@ -672,6 +713,17 @@ func (st *driverStmt) exec(ctx context.Context, args []driver.NamedValue) (*stmt
 	if err != nil {
 		return nil, fmt.Errorf("trino: %v", err)
 	}
+
+	// init query callback
```
losipiuk, line 717: I think more appropriate comment would be
// initial progress callback call

trino/trino.go
```diff
@@ -1649,3 +1725,21 @@ func (s *NullSlice3Map) Scan(value interface{}) error {
...
 }
+
+// QueryResponse object used for callback
+type QueryResponse struct {
+	ID         string    `json:"id"`
+	QueryStats stmtStats `json:"stmtStats"`
+	Error      stmtError `json:"error"`
+}
+
+type QueryCallbackPeriod struct {
```
losipiuk, line 1736: this should be private

trino/trino.go
```diff
@@ -1649,3 +1725,21 @@ func (s *NullSlice3Map) Scan(value interface{}) error {
...
+	Period    float64   `json:"period"`
+	CheckTime time.Time `json:"checkTime"`
```
losipiuk, line 1738: CheckTime -> lastCallbackTime

trino/trino.go
```diff
@@ -1649,3 +1725,21 @@ func (s *NullSlice3Map) Scan(value interface{}) error {
...
+
+// QueryResponse object used for callback
+type QueryResponse struct {
+	ID         string    `json:"id"`
+	QueryStats stmtStats `json:"stmtStats"`
+	Error      stmtError `json:"error"`
+}
+
+type QueryCallbackPeriod struct {
+	Period    float64   `json:"period"`
```
losipiuk, line 1737: use `Duration` instead of `float64`

losipiuk (commented):
Thanks. Skimmed.

Note that I rebased and force-pushed to PR branch.

trino/trino.go
```diff
@@ -99,26 +99,31 @@ var (
...
+	trinoSetCatalogHeader    = "X-Trino-Set-Catalog"
+	trinoSetSchemaHeader     = "X-Trino-Set-Schema"
+	trinoSetPathHeader       = "X-Trino-Set-Path"
+	trinoSetSessionHeader    = "X-Trino-Set-Session"
+	trinoClearSessionHeader  = "X-Trino-Clear-Session"
+	trinoSetRoleHeader       = "X-Trino-Set-Role"
+	trinoQueryCallbackHeader = "X-Query-Stats-Callback"
+	trinoQueryCallbackPeriod = "X-Query-Stats-CallbackPeriod"
+
+	// KerberosEnabledConfig KerberosEnabled
```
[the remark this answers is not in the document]
Chaho12: Will do remove it. 
I added the comments because my Goland IDE gives warnings that exported variables should have comments.

Chaho12:
@losipiuk Sorry for late update, I was busy with my companies work. 
I rebased and updated code to latest style.

trino/trino.go
Chaho12: This is sample code that we are current using in our project (mind taht name is out-of-date, since code is not updated yet)

Since query response has QueryStats we can access data inside within.
```go
type QueryCallback struct {
	target *TrinoResultWithProgress
	Ctx    *ctx.Context
}

// Function called for every request made to Query
func (qc *QueryCallback) StatsUpdated(qr trino.QueryResponse) {
	percentage := float32(math.Min(100.0, float64(qr.QueryStats.CompletedSplits*100.0)/float64(qr.QueryStats.TotalSplits)))
	if qr.QueryStats.TotalSplits == 0 || qc.target.Percentage > percentage {
		return
	}
	now := time.Now()
	qc.target.QueryID = qr.ID
	qc.target.State = qr.QueryStats.State
	qc.target.Scheduled = qr.QueryStats.Scheduled
	qc.target.Nodes = qr.QueryStats.Nodes
	qc.target.TotalSplits = qr.QueryStats.TotalSplits
	qc.target.QueuedSplits = qr.QueryStats.QueuesSplits
	qc.target.CompletedSplits = qr.QueryStats.CompletedSplits
	qc.target.ProcessedRows = qr.QueryStats.ProcessedRows
	qc.target.ProcessedBytes = qr.QueryStats.ProcessedBytes
	qc.target.Percentage = percentage
	qc.target.CheckTime = now
	if _, err := qc.Ctx.MainOrm.ID(qc.target.ParamHash).
		Omit("start_time", "end_time", "logs", "fetched_line", "fetched_path").Update(*qc.target); err != nil {
		util.LogWarn("can't update trino state at trino query history table : %+v, because %+v", *qc.target, err)
	}
}
```

trino/trino.go
```diff
@@ -104,6 +104,7 @@ var (
 
 const (
 	trinoHeaderPrefix = `X-Trino-`
+	trinoQueryPrefix  = `X-Query-`
```
nineinchnick, line 107: We already recognize `X-Trino-User` in named parameters. I don't you should introduce a new prefix.

trino/trino.go
```diff
@@ -595,6 +601,23 @@ func (st *driverStmt) CheckNamedValue(arg *driver.NamedValue) error {
 	if reflect.TypeOf(arg.Value).Kind() == reflect.Slice {
 		return nil
 	}
+
+	// check callback
+	if arg.Name == trinoProgressCallbackParam {
+		callback, ok := arg.Value.(QueryProgressCallback)
+		if ok {
+			st.conn.callback = callback
```
nineinchnick, line 609: Why do this both here and in `exec()`? I think this method should only do validation and not modify any properties of the connection.

trino/trino.go
```diff
@@ -976,11 +1022,34 @@ func (qr *driverRows) fetch(allowEOF bool) error {
...
+
 	if len(qr.data) == 0 {
 		if qr.nextURI != "" {
+			if qr.stmt.conn.callback != nil {
+				if qrStats.QueryStats.State == "RUNNING" {
+					currentTime := time.Now()
+					diff := currentTime.Sub(qr.stmt.conn.callbackPeriod.LastCallbackTime).Milliseconds()
+					period := qr.stmt.conn.callbackPeriod.Period.Milliseconds()
+					// Check if period has passed
+					if diff >= period {
```
nineinchnick, line 1040: Why you cast it to milliseconds? Isn't `time.Time` comparable?

trino/trino.go
```diff
@@ -976,11 +1022,34 @@ func (qr *driverRows) fetch(allowEOF bool) error {
...
+						qr.stmt.conn.callback.ProgressUpdated(qrStats)
```
nineinchnick, line 1041: Can this be blocking? I think instead of calling it here, you should write to a (buffered?) channel and have a goroutine reading from it and calling `ProgressUpdated`. If the channel is full, maybe it would be ok to drop updates.

trino/trino.go
```diff
@@ -1943,3 +2012,19 @@ func (s *NullSlice3Map) Scan(value interface{}) error {
...
+	QueryStats stmtStats `json:"stmtStats"`
+	Error      stmtError `json:"error"`
+}
+
+type queryProgressCallbackPeriod struct {
+	Period           time.Duration `json:"period"`
+	LastCallbackTime time.Time     `json:"lastCallbackTime"`
+}
+
+type QueryProgressCallback interface {
```
nineinchnick, line 2027: It would be idiomatic in Go to call this `ProgressUpdater` and have a method called `Update`.

trino/trino_test.go
```diff
@@ -249,6 +249,82 @@ func TestQueryForUsername(t *testing.T) {
 	}
 }
 
+type TestQueryProgressCallback struct{}
+
+func (*TestQueryProgressCallback) ProgressUpdated(qr QueryProgressCallbackResponse) {
+	fmt.Printf("%+v, %+v \n", qr.QueryStats.State, qr.QueryStats.ProgressPercentage)
```
nineinchnick, line 255: You shouldn't print anything here. This implies you rely on a human to do the assert. Instead, you should increment some counter, and do a regular assert of its value at the end of the test.

trino/trino_test.go
```diff
@@ -249,6 +249,82 @@ func TestQueryForUsername(t *testing.T) {
...
+
+	db, err := sql.Open("trino", dsn)
+	require.NoError(t, err)
+
+	t.Cleanup(func() {
+		assert.NoError(t, db.Close())
+	})
+
+	callback := &TestQueryProgressCallback{}
+	callbackPeriod, err := time.ParseDuration("100ms")
```
nineinchnick, line 311: Same here, you should actually assert that the callback does not get called too often/early. Maybe instead of just a counter, the `TestQueryProgressCallback` should track the minimum interval since the last update. The period should be much smaller so the test can finish quicker.

trino/trino.go
```diff
@@ -1943,3 +2012,19 @@ func (s *NullSlice3Map) Scan(value interface{}) error {
 	s.Valid = true
 	return nil
 }
+
+type QueryProgressCallbackResponse struct {
+	ID         string    `json:"id"`
+	QueryStats stmtStats `json:"stmtStats"`
+	Error      stmtError `json:"error"`
```
nineinchnick, line 2019: Having an error field in it implies users might try to use it for error handling instead of the regular flow of `database/sql`. Isn't the `QueryStats.State` field enough?

trino/trino.go
```diff
@@ -976,11 +1022,34 @@ func (qr *driverRows) fetch(allowEOF bool) error {
...
+	qrStats := QueryProgressCallbackResponse{
+		ID:         qresp.ID,
+		QueryStats: qresp.Stats,
+		Error:      qresp.Error,
+	}
+
 	if len(qr.data) == 0 {
 		if qr.nextURI != "" {
+			if qr.stmt.conn.callback != nil {
+				if qrStats.QueryStats.State == "RUNNING" {
```
nineinchnick, line 1035: Why only `RUNNING`?

trino/trino.go
```diff
@@ -976,11 +1022,34 @@ func (qr *driverRows) fetch(allowEOF bool) error {
...
+					currentTime := time.Now()
```
nineinchnick, line 1036: This should be a separate method.

trino/trino.go
```diff
@@ -976,11 +1022,34 @@ func (qr *driverRows) fetch(allowEOF bool) error {
...
+
+	qrStats := QueryProgressCallbackResponse{
+		ID:         qresp.ID,
+		QueryStats: qresp.Stats,
+		Error:      qresp.Error,
+	}
+
 	if len(qr.data) == 0 {
 		if qr.nextURI != "" {
+			if qr.stmt.conn.callback != nil {
```
nineinchnick, line 1034: The body of this condition should be a separate method.

nineinchnick: Why this is only called when there's no data?

trino/trino.go
Chaho12: Ok, then would `X-Trino-Progress-Callback` and `X-Trino-Progress-Callback-Period` be ok?

nineinchnick:
@Chaho12 can you address previous comments too? If you think they're resolved, mark them as such. If you disagree with them, please explain why. There are a lot of them that don't have any response and are still valid.

trino/trino.go
Chaho12: Oops, my bad. I forgot to remove it at exec() part because `CheckNamedValue` is called by sql.

trino/trino.go
Chaho12: I i don't quite understand how to with simply using `time.Time` comparable since time.Sub returns time.Duration.
1. need to find time elapsed since last call back time
2. compare with period
3. if elapsed time exceeds period, update progress

trino/trino.go
nineinchnick: It should be done in exec, even if just for consistency with `X-Trino-User`.

trino/trino.go
nineinchnick: Oh sorry, I meant comparing `time.Duration`. You don't need to extract milliseconds. This program:
```go
package main

import (
	"fmt"
	"time"
)

func main() {
	now := time.Now()
	before := time.Date(2009, 11, 10, 21, 0, 0, 0, time.UTC)
	fmt.Printf("now: %s,\nbefore: %s\n", now, before)
	diff := now.Sub(before)
	period := 1 * time.Hour
	if diff > period {
		fmt.Printf("%s > %s\n", diff, period)
	} else {
		fmt.Printf("%s <= %s\n", diff, period)
	}
}
```
prints:
```
now: 2009-11-10 23:00:00 +0000 UTC m=+0.000000001,
before: 2009-11-10 21:00:00 +0000 UTC
2h0m0s > 1h0m0s
```
https://go.dev/play/p/Fsi0p4YGKNx

trino/trino.go
Chaho12: Will fix it as you said. 
Also, what do you think about adding default period if user does not give period header? 1000ms?

trino/trino.go
Chaho12: Hmm, I haven't had a thought on this that it could be blocking (depends on the ProgressUpdated that user defines, and period that user set ?)
Thx for insight. i'll try it out using channel.

trino/trino.go
nineinchnick: I don't think we can find a good default value, so let's just make it required.

nineinchnick: Ah, I forgot that only primitive types are supported as query params by default, so if a callback is passed we'd still need to check for it here. But I'd still separate validation from the assignment.

Chaho12: I can't seem to find appropriate location for required header, please let me know.

For now, I put it at exec() function, after all headers.
image

trino/trino.go
nineinchnick: If you'll have any issues implementing it I could help, we can sync on the Trino Slack.

trino/trino_test.go
Chaho12: how about this? check for queue and finished as it is always called.
```go
type TestQueryProgressCallback struct {
	statusMap map[string]float32
}

func (qpc *TestQueryProgressCallback) Update(qr QueryProgressCallbackResponse) {
	qpc.statusMap[qr.QueryStats.State] = qr.QueryStats.ProgressPercentage
}
...

assert.Contains(t, statusMap, "QUEUED")
assert.Contains(t, statusMap, "FINISHED")

File: trino/trino.go (hunk 18 above)
Comment (Chaho12) [reply trino/trino.go:2019]: Is it weird to use it for error logging/handling? It has detailed information on what error happened, so it was useful when analyzing cause of the error.

Here is part of my code that I am using to log error.
```go
func (qc *QueryCallback) StatsUpdated(qr trino.QueryResponse) {
    ...
    if op, err := NewTrinoOperationProgress(target, &qr); err == nil {
        log := TrinoLogJson{
	      StatusCode:   op.Status,
	      ErrorCode:    "N/A",
	      ErrorMessage: "Trino Operation Progress",
	      InfoMessage:  qr.Error.Message,
	}
	err = op.Update(qc.Ctx, log)
	...
    }
}
```

And here is what gets logged in mysql table
```
{"StatusCode":"TABLE_NOT_FOUND","ErrorCode":"46","ErrorMessage":"io.trino.spi.TrinoException error - [line 3:3: Table '.......' does not exist]","InfoMessage":"[io.trino.sql.analyzer.SemanticExceptions.semanticException(SemanticExceptions.java:48) io.trino.sql.analyzer.SemanticExceptions.semanticException(SemanticExceptions.java:43) io.trino.sql.analyzer.StatementAnalyzer$Visitor.visitTable(StatementAnalyzer.java:1243) io.trino.sql.analyzer.StatementAnalyzer$Visitor.visitTable(StatementAnalyzer.java:343) io.trino.sql.tree.Table.accept(Table.java:53) io.trino.sql.tree.AstVisitor.process(AstVisitor.java:27) io.trino.sql.analyzer.StatementAnalyzer$Visitor.process(StatementAnalyzer.java:360) io.trino.sql.analyzer.StatementAnalyzer$Visitor.analyzeFrom(StatementAnalyzer.java:2770) io.trino.sql.analyzer.StatementAnalyzer$Visitor.visitQuerySpecification(StatementAnalyzer.java:1589) io.trino.sql.analyzer.StatementAnalyzer$Visitor.visitQuerySpecification(StatementAnalyzer.java:343) io.trino.sql.tree.QuerySpecification.accept(QuerySpecification.java:155) io.trino.sql.tree.AstVisitor.process(AstVisitor.java:27) io.trino.sql.analyzer.StatementAnalyzer$Visitor.process(StatementAnalyzer.java:360) io.trino.sql.analyzer.StatementAnalyzer$Visitor.process(StatementAnalyzer.java:370) io.trino.sql.analyzer.StatementAnalyzer$Visitor.visitQuery(StatementAnalyzer.java:1075) io.trino.sql.analyzer.StatementAnalyzer$Visitor.visitQuery(StatementAnalyzer.java:343) io.trino.sql.tree.Query.accept(Query.java:107) io.trino.sql.tree.AstVisitor.process(AstVisitor.java:27) io.trino.sql.analyzer.StatementAnalyzer$Visitor.process(StatementAnalyzer.java:360) io.trino.sql.analyzer.StatementAnalyzer.analyze(StatementAnalyzer.java:323) io.trino.sql.analyzer.Analyzer.analyze(Analyzer.java:91) io.trino.sql.analyzer.Analyzer.analyze(Analyzer.java:83) io.trino.execution.SqlQueryExecution.analyze(SqlQueryExecution.java:263) io.trino.execution.SqlQueryExecution.\u003cinit\u003e(SqlQueryExecution.java:186) io.trino.execution.SqlQueryExecution$SqlQueryExecutionFactory.createQueryExecution(SqlQueryExecution.java:768) io.trino.dispatcher.LocalDispatchQueryFactory.lambda$createDispatchQuery$0(LocalDispatchQueryFactory.java:129) io.trino.$gen.Trino_353____20210715_095735_2.call(Unknown Source) com.google.common.util.concurrent.TrustedListenableFutureTask$TrustedFutureInterruptibleTask.runInterruptibly(TrustedListenableFutureTask.java:125) com.google.common.util.concurrent.InterruptibleTask.run(InterruptibleTask.java:69) com.google.common.util.concurrent.TrustedListenableFutureTask.run(TrustedListenableFutureTask.java:78) java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1130) java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:630) java.base/java.lang.Thread.run(Thread.java:832)]"}
```

trino/trino.go
Chaho12: I thought it would be meaningless to make further request on other states since
1. we already make initial progress callback call in beginning
2. make a call when query is finished.

So, at minimum state updated is (QUEUED, FINISHED)

trino/trino.go
Chaho12: It maybe due to my lack of understanding trino rest api and go code, 
but I thought that running state is when data is being processed and not available to user yet.

trino/trino.go
Chaho12: I greatly appreciate on giving detailed feedback and pointing out places where it could lead to problems. I am learning a lot :)

I'll first try out on using channel and share to code with the rest of changes I made.

trino/trino_test.go
nineinchnick: That doesn't assert if the period is used at all.

trino/trino.go
nineinchnick: You shouldn't do error handling when checking query progress, but when reading query results. If Trino returns an error, you'll be able to read it after calling `DB.ExecContext()`, `DB.QueryContext()`. If the error happens during reading results using `Rows.Next()`, you'd get it from `Rows.Err()`, which you should always call after `Rows.Next()` returns false.

trino/trino.go
nineinchnick: The decision on how to process the state should be left to the progress callback.

trino/trino.go
nineinchnick: That's an implementation detail. This should be always called when there's a response from the server.

trino/trino.go
nineinchnick: Being able to do error handling in two different places would be confusing and error-prone.

nineinchnick:
@Chaho12 you marked most comments as resolved, but I don't see them being addressed. Did you forget to push your latest changes? Please resolve comments only after pushing changes, otherwise, another reviewer might add similar comments.

trino/trino.go
Chaho12: `error handling when checking query progress, but when reading query results` Got it thx :)

trino/trino.go
Chaho12: As of now, I changed so that changed so that updates when there is change in any state OR if callback period has passed. :)
```go
func (qr *driverRows) updateCallback(qrStats QueryProgressCallbackResponse) {
	currentTime := time.Now()
	diff := currentTime.Sub(qr.stmt.conn.callbackPeriod.LastCallbackTime)
	period := qr.stmt.conn.callbackPeriod.Period
	// Check if period has passed OR if query state changes
	if diff >= period || qrStats.LastQueryState != qrStats.QueryStats.State {
		qr.stmt.conn.callback.Update(qrStats)
		qr.stmt.conn.callbackPeriod.LastCallbackTime = currentTime
	}
}
```

trino/trino.go
Chaho12: I left comment below as it seems to be similar topic
- https://github.com/trinodb/trino-go-client/pull/14#discussion_r937292355

trino/trino.go
Chaho12: Removed it, and added field called lastqueryState
```go
type QueryProgressCallbackResponse struct {
	ID             string
	QueryStats     stmtStats
	LastQueryState string
}

File: trino/trino_test.go (hunk 17 above)
Comment (Chaho12) [reply trino/trino_test.go]: If you have time, PTAL :)

File: trino/trino.go (hunk 21 above)
Comment (Chaho12) [reply trino/trino.go:1034]: BTW am working on change this to channel (am keep getting interrupts so it is taking some time) i'll let you know when am done with it

File: trino/trino.go (hunk 14 above)
Comment (Chaho12) [reply trino/trino.go]: What do you think of this change?
https://github.com/trinodb/trino-go-client/pull/14/commits/fa70b40b593afc0c49377687be1e667a70ccb79a

I wonder if it is allright to call go routine like this (simply open channel and send data if conn has callback)
```go
c := make(chan QueryProgressCallbackResponse)
go qr.updateCallback(c)
	if len(qr.data) == 0 {
		if qr.stmt.conn.callback != nil {
			c <- qrStats
		}
		if qr.nextURI != "" {
			return qr.fetch(allowEOF)
		}
		if allowEOF {
			qr.err = io.EOF
			return qr.err
		}
	}
	...
	qr.rowsAffected = qresp.UpdateCount
	if qr.stmt.conn.callback != nil {
		c <- qrStats
	}
```

nineinchnick: You'll need to:
1. Create two unbuffered channels when the query begins in `exec()`, one for query stats and another one to signal when the query is done. The second one should be of an empty type `struct{}`, because it won't transmit data, only be closed.
2. Start a goroutine that would read from the channel and call the update callback. It's important to control when this goroutine should end (when the query is done). That's why we need the second channel.
3. When receiving a query response, send query stats into the channel. Ignore the error when the channel is full.
4. When the query finishes, close the stop channel and the stats channel (in this order).

An example goroutine:
```go
done := make(chan struct{})
go func() {
  for {
    select {
    case stats := <-statsCh:
      update(stats)
    case <-done:
      return
    }
  }
}()
```

Sending to a channel but ignoring when it's full:
```go
select {
    case statsCh <- stats:
    default:
        // ignore when can't send stats
    }
```

trino/trino.go
```diff
@@ -1043,6 +1105,18 @@ func (qr *driverRows) initColumns(qresp *queryResponse) error {
 	return nil
 }
 
+func (qr *driverRows) updateCallback(c chan QueryProgressCallbackResponse) {
+	qrStats := <-c
+	currentTime := time.Now()
+	diff := currentTime.Sub(qr.stmt.conn.callbackPeriod.LastCallbackTime)
+	period := qr.stmt.conn.callbackPeriod.Period
+	// Check if period has passed OR if query state changes
+	if diff >= period || qrStats.LastQueryState != qrStats.QueryStats.State {
```
nineinchnick, line 1114: Comparing if the state changed shouldn't be necessary. Stats include query progress, so you'd update the progress multiple times when the state is `RUNNING` and doesn't change.

Chaho12: I included comparing state because if period hasn't passed, and state changes from RUNNING -> FINISHING/FINISHED and time take for finishing was less than period given,
last callback response would be running (does not return finish).

nineinchnick: I totally missed the comment, sorry!

trino/trino.go
Chaho12: Quick question, why channel at exec and not at fetch? I thought exec is only called once, so need to worry for blocking.

nineinchnick: You should do the setup once, in exec, when the query starts. So create all required channels and set up the goroutine once, instead of creating goroutines for every HTTP request. This will help avoid issues with memory usage. It's equally important to ensure the goroutine will finish when the query is done, to avoid leaking it.

trino/trino.go
```diff
@@ -693,6 +712,19 @@ func (st *driverStmt) exec(ctx context.Context, args []driver.NamedValue) (*stmt
 		hs = make(http.Header)
 		var ss []string
 		for _, arg := range args {
+			if strings.Contains(arg.Name, "Progress-Callback") {
```
nineinchnick, line 715: This condition is unnecessary, since you test for exact names in the following condition blocks.

trino/trino.go
```diff
@@ -1943,3 +2017,19 @@ func (s *NullSlice3Map) Scan(value interface{}) error {
...
+	LastQueryState string
+}
+
+type queryProgressCallbackPeriod struct {
+	Period           time.Duration
+	LastCallbackTime time.Time
+}
+
+type ProgressUpdater interface {
+	// Update : Called on progress update, before and after query execution finished.
```
nineinchnick, line 2033: Nit: the comment should start with the function name but it should be a valid English sentence, so use something like:
```suggestion
	// Update the query progress, immediately when the query starts, when receiving data, and once when the query is finished.
```

trino/trino.go
Chaho12: I tried to do as you said, but how do I call statsCh inside fetch function? I can't pass it as parameter so statsCh variable does not exist in fetch function scope.

nineinchnick: Add fields to the statement and driverRows structs.

Chaho12:
@nineinchnick Once again, I greatly appreciate your time to help me out. I am learning a lot from your comments! PTAL on changes :)

Chaho12:
@losipiuk I believe I made changes to all the points noted. PTAL when you've got time :) thx for giving feedbacks.

CONTRIBUTING.md
```diff
@@ -8,6 +8,14 @@ In order to accept your pull request, we need you to [submit a CLA](https://gith
 
 By contributing to Trino, you agree that your contributions will be licensed under the [Apache License Version 2.0 (APLv2)](LICENSE).
 
+# Go Test
```
nineinchnick, line 11: This is not related to other changes in the PR. Can you squash all commits but keep this in a separate one, to make reviewing easier?

trino/trino.go
```diff
@@ -576,11 +586,14 @@ func (st *driverStmt) ExecContext(ctx context.Context, args []driver.NamedValue)
...
 		nextURI:      sr.NextURI,
 		rowsAffected: sr.UpdateCount,
+		statsCh:      st.statsCh,
+		doneCh:       st.doneCh,
 	}
 	// consume all results, if there are any
 	for err == nil {
 		err = rows.fetch(true)
 	}
+	st.doneCh <- struct{}{}
```
nineinchnick, line 596: You should close this channel here.

trino/trino.go
```diff
@@ -678,21 +699,38 @@ func (st *driverStmt) QueryContext(ctx context.Context, args []driver.NamedValue
 		stmt:    st,
 		queryID: sr.ID,
 		nextURI: sr.NextURI,
+		statsCh: st.statsCh,
+		doneCh:  st.doneCh,
 	}
 	if err = rows.fetch(false); err != nil {
 		return nil, err
 	}
+	st.doneCh <- struct{}{}
```
nineinchnick, line 708: Close this channel.

trino/trino.go
```diff
@@ -739,6 +780,32 @@ func (st *driverStmt) exec(ctx context.Context, args []driver.NamedValue) (*stmt
 	if err != nil {
 		return nil, fmt.Errorf("trino: %v", err)
 	}
+
+	// initialize callback go func
+	go func() {
```
nineinchnick, line 785: This should only be done if `st.conn.callback != nil`

trino/trino.go
```diff
@@ -678,21 +699,38 @@ func (st *driverStmt) QueryContext(ctx context.Context, args []driver.NamedValue
 		stmt:    st,
 		queryID: sr.ID,
 		nextURI: sr.NextURI,
+		statsCh: st.statsCh,
+		doneCh:  st.doneCh,
```
nineinchnick, line 703: The done channel is not used in `driverRows`

trino/trino.go
```diff
@@ -1043,6 +1125,22 @@ func (qr *driverRows) initColumns(qresp *queryResponse) error {
 	return nil
 }
 
+func (qr *driverRows) updateCallback(qrStats QueryProgressCallbackResponse) {
+	currentTime := time.Now()
+	diff := currentTime.Sub(qr.stmt.conn.callbackPeriod.LastCallbackTime)
+	period := qr.stmt.conn.callbackPeriod.Period
+	// Check if period has passed OR if query state changes
+	if diff >= period || qr.stmt.conn.callbackPeriod.LastQueryState != qrStats.QueryStats.State {
```
nineinchnick, line 1133: Reverse this condition and do an empty return here to avoid one level of indentation of the remaining code.

trino/trino.go
```diff
@@ -976,7 +1046,16 @@ func (qr *driverRows) fetch(allowEOF bool) error {
...
 	qr.data = qresp.Data
 	qr.nextURI = qresp.NextURI
+
+	qrStats := QueryProgressCallbackResponse{
+		ID:         qresp.ID,
+		QueryStats: qresp.Stats,
+	}
+
 	if len(qr.data) == 0 {
+		if qr.stmt.conn.callback != nil {
```
nineinchnick, line 1056: You could move this check into the `updateCallback` function. You could also pass a reference to the whole response there and it should create the `QueryProgressCallbackResponse` only when necessary.

nineinchnick (commented):
Almost there, looks good!

trino/trino.go
Chaho12: added `close(st.doneCh)` on line after this.

trino/trino.go
Chaho12: added `close(st.doneCh)` on line after this.

trino/trino.go
nineinchnick: Just close it without sending anything. You can't receive an empty struct anyway, so closing is equivalent and it can be done only once.

trino/trino.go
```diff
@@ -1043,6 +1115,30 @@ func (qr *driverRows) initColumns(qresp *queryResponse) error {
 	return nil
 }
 
+func (qr *driverRows) updateCallback(id string, stats stmtStats) {
+	if qr.stmt.conn.callback != nil {
```
nineinchnick, line 1119: Reverse the condition and do an early return. This allows to avoid one level of indentation, making the whole function more readable

Chaho12: Wow... I have never thought of indentation level so at first I did not understand properly on what you meant by reversing the condition. I am grateful for spotting this out.

nineinchnick: This explains it better: https://go.dev/doc/effective_go#if

trino/trino_test.go
```diff
@@ -249,6 +250,99 @@ func TestQueryForUsername(t *testing.T) {
...
+		assert.Equal(t, "2", ts, "Expected value does not equal result value")
+	}
+
+	if err = rows.Err(); err != nil {
+		t.Fatal(err)
+	}
+
+	// sort time in order to calculate interval
+	var keys []time.Time
+	for k := range statusMap {
```
nineinchnick, line 332: Tests are failing now with a race condition here. The race detector sees two goroutines that read and write from the same map and it cannot determine if the goroute that calls the update callback has finished or not. To eliminate this possible race, you could close the `statsCh` channel before exiting the goroutine and try to read from it after closing `doneCh` in `ExecContext` and `QueryContext` to make sure it has been closed.

trino/trino.go
```diff
@@ -100,6 +100,9 @@ var (
 
 	// ErrInvalidResponseType indicates that the server returned an invalid type definition.
 	ErrInvalidResponseType = errors.New("trino: server response contains an invalid type")
+
+	// ErrInvalidProgressCallbackHeader indicates that server did not get valid headers for progress callback
+	ErrInvalidProgressCallbackHeader = errors.New("trino: " + trinoProgressCallbackParam + trinoProgressCallbackPeriodParam + " both must be set when using progress callback")
```
losipiuk, line 105: nit: `"trino: both " + trinoProgressCallbackParam + " and " + trinoProgressCallbackPeriodParam + " must be set when using progress callback"`

trino/trino.go
```diff
@@ -1943,3 +2039,19 @@ func (s *NullSlice3Map) Scan(value interface{}) error {
 	s.Valid = true
 	return nil
 }
+
+type QueryProgressCallbackResponse struct {
+	ID         string
```
losipiuk, line 2045: `ID` -> `QueryId`

trino/trino.go
```diff
@@ -1943,3 +2039,19 @@ func (s *NullSlice3Map) Scan(value interface{}) error {
 	s.Valid = true
 	return nil
 }
+
+type QueryProgressCallbackResponse struct {
```
losipiuk, line 2044: I think `QueryProgressInfo` (or sth similar) would be better name.

trino/trino.go
```diff
@@ -1943,3 +2039,19 @@ func (s *NullSlice3Map) Scan(value interface{}) error {
...
+	QueryStats stmtStats
+}
+
+type queryProgressCallbackPeriod struct {
+	Period           time.Duration
+	LastCallbackTime time.Time
+	LastQueryState   string
+}
+
+type ProgressUpdater interface {
```
losipiuk, line 2055: maybe `QueryProgressCallback`?

nineinchnick: `ProgressUpdater` is idiomatic: https://go.dev/doc/effective_go#interface-names

trino/trino.go
```diff
@@ -1043,6 +1115,30 @@ func (qr *driverRows) initColumns(qresp *queryResponse) error {
...
+		qrStats := QueryProgressCallbackResponse{
+			ID:         id,
+			QueryStats: stats,
+		}
+		currentTime := time.Now()
+		diff := currentTime.Sub(qr.stmt.conn.callbackPeriod.LastCallbackTime)
+		period := qr.stmt.conn.callbackPeriod.Period
+
+		// Check if query state changes OR if period has passed
+		if qr.stmt.conn.callbackPeriod.LastQueryState != qrStats.QueryStats.State || diff >= period {
```
losipiuk: Why do you think `qr.stmt.conn.callbackPeriod.LastQueryState != qrStats.QueryStats.State` is valuable? I would inform even if query is stalling.  This is also a piece of information.

nineinchnick: This condition triggers an update when there _was_ a state change but not enough time passed since the last update.

trino/trino.go
```diff
@@ -1043,6 +1115,30 @@ func (qr *driverRows) initColumns(qresp *queryResponse) error {
 	return nil
 }
 
+func (qr *driverRows) updateCallback(id string, stats stmtStats) {
```
losipiuk, line 1116: would `scheduleQueryProgressCallbackIfNeeded` be a better name here?

trino/trino.go
```diff
@@ -739,6 +780,32 @@ func (st *driverStmt) exec(ctx context.Context, args []driver.NamedValue) (*stmt
 	if err != nil {
 		return nil, fmt.Errorf("trino: %v", err)
 	}
+
+	// initial progress callback call
```
losipiuk, line 782: Why do we need special code to handle initial call?

trino/trino.go
Chaho12: Hmm... seems to be quite long for function name, so how about `scheduleProgressUpdate` or `updateProgress`?

trino/trino.go
Chaho12: Because it takes small amount of time before rows.fetch is executed, some short queries might end too fast with first callback with `FINISHED` state even if callback period is set fast. This seems to be weird as all intermediate steps is skipped, so I added initial call to update as soon as we submitted the query.

trino/trino_test.go
Chaho12: While I was trying to close the `statsCh` channel inside the go routine, I noticed that `rows.Next()` calls fetch function, which calls updateCallback function.
So I added code to check and return if `statsCh` channel is close. Did I understood your intention correctly?
- https://github.com/trinodb/trino-go-client/pull/14/commits/f2af3433d9581a7cd1b744754b5ff8845ea4cb52#diff-0142aac4be66432bab473dd519af9593a22e6842a9a15a5251c14d97b3208519R791

trino/trino.go
losipiuk: `scheduleProgressUpdate` is fine

trino/trino.go
losipiuk: Makes sense.

Should we maybe then call fields in `Conn`
```
	progressUpdater        ProgressUpdater
	progressUpdaterPeriod  queryProgressCallbackPeriod
```
instead:
```
	callback        ProgressUpdater
	callbackPeriod  queryProgressCallbackPeriod
```

trino/trino.go
```diff
@@ -678,21 +698,37 @@ func (st *driverStmt) QueryContext(ctx context.Context, args []driver.NamedValue
...
 		for _, arg := range args {
+			if arg.Name == trinoProgressCallbackParam {
+				callback := arg.Value.(ProgressUpdater)
+				st.conn.callback = callback
+				continue
+			}
+			if arg.Name == trinoProgressCallbackPeriodParam {
+				period := arg.Value.(time.Duration)
+				st.conn.callbackPeriod.Period = period
+				st.conn.callbackPeriod.LastCallbackTime = time.Now()
```
losipiuk, line 728: nit: more natural place to set this would be with code which does initial call to the callback.

trino/trino.go
```diff
@@ -678,21 +698,37 @@ func (st *driverStmt) QueryContext(ctx context.Context, args []driver.NamedValue
...
+	st.doneCh = make(chan struct{})
 	query := st.query
 	var hs http.Header
 
 	if len(args) > 0 {
 		hs = make(http.Header)
 		var ss []string
 		for _, arg := range args {
+			if arg.Name == trinoProgressCallbackParam {
+				callback := arg.Value.(ProgressUpdater)
```
losipiuk, line 721: Do you need `callback` and `period` variables, or you can just inline those and assign directly to `st.conn.callback`st.conn.callbackPeriod.Period`?
 I recall there was some go weirdness that you sometimes needed to copy iteration variable into new one - but I do not think this is the case.

trino/trino.go
losipiuk: fair. 

Yet please move the comment to be above 
```
srStats := QueryProgressInfo{
			QueryId:    sr.ID,
			QueryStats: sr.Stats,
		}
```

currenntly it is mispositioned.

trino/trino.go
Chaho12: changed field names :)

trino/trino_test.go
nineinchnick: That should not be necessary, but it means that `doneCh` is closed prematurely. I overlooked this, it should be closed in `func (qr *driverRows) Close() error`. `Next()` should never be called after `Close()`.

trino/trino.go
```diff
@@ -1043,6 +1114,40 @@ func (qr *driverRows) initColumns(qresp *queryResponse) error {
...
 }
 
+func (qr *driverRows) updateCallback(id string, stats stmtStats) {
+	if qr.stmt.conn.callback == nil {
+		return
+	}
+
+	// Check and return if statsCh Channel is closed
+	select {
+	case <-qr.statsCh:
```
losipiuk, line 1123: don't you read and drop value from `statsCh` with this statement if this is not closed?

trino/trino_test.go
Chaho12: How about this? calling done channel to close go routine callback channels
```go
func (qr *driverRows) Close() error {
	if qr.stmt.conn.progressUpdater != nil {
		qr.doneCh <- struct{}{}
	}
```
with go function as 
```go
		go func() {
			for {
				select {
				case stats := <-st.statsCh:
					st.conn.progressUpdater.Update(stats)
				case <-st.doneCh:
					close(st.statsCh)
					close(st.doneCh)
					return
				}
			}
		}()

```

trino/trino.go
Chaho12: Even if this code is removed, thx for pointint it out.
- https://github.com/trinodb/trino-go-client/pull/14#discussion_r949984630

I tested it at [playground](https://go.dev/play/p/5arKW_2EQJ3) and very interestingly `poing` disappears, however, it appears if I uncomment line 34 and send another message `messages <- "new pong"`
```
// before uncomment
ping
pong
before open
where is my pong?

...
//after uncomment
ping
before open
where is my pong?
pong
new pong

File: trino/trino_test.go (hunk 33 above)
Comment (nineinchnick) [reply trino/trino_test.go:332]: You can't close a channel twice. If the `case <-st.doneCh` branch executes, this channel should be already closed. And you can't receive anything on an empty struct channel (because an empty struct has a size of 0, so it's not even sent to the channel).

Comment (nineinchnick) [reply trino/trino_test.go:332]: So the `doneCh` channel should be closed in the `rows.Close()` method. It would be best if all channels were uni-directional, and closing a channel is equal to writing to it, so you shouldn't close it where you read from it because that would mean it has to be bi-directional. Closing `statsCh` in the goroutine is an exception, we could avoid that by introducing yet another channel, but I don't think we have to right now.

Comment (Chaho12) [reply trino/trino_test.go:332]: I don't quite understand when you mean by `And you can't receive anything on an empty struct channel (because an empty struct has a size of 0, so it's not even sent to the channel).` because on playground, when calling `close(done)` in line 27 or calling `done <- struct{}{}` in line 26 results the same and sending to empty struct channel works.
- https://go.dev/play/p/L91TX9ngoNI

If I simply remove `close(st.doneCh)`, i do not get any errors, `however`, I get race condition error if i move `close(st.doneCh)` and replace it at `qr.doneCh <- struct{}{}`
image

```
==================
WARNING: DATA RACE
Read at 0x00c00028eba0 by goroutine 46:
  runtime.mapdelete()
      /usr/local/go/src/runtime/map.go:685 +0x49c
  github.com/trinodb/trino-go-client/trino.TestQueryProgressWithCallbackPeriod()
      /Users/user/work/trino-go-client/trino/trino_test.go:332 +0x7f0
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1259 +0x198

Previous write at 0x00c00028eba0 by goroutine 47:
  runtime.mapaccessK()
      /usr/local/go/src/runtime/map.go:511 +0x1ec
  github.com/trinodb/trino-go-client/trino.(*TestQueryProgressCallback).Update()
      /Users/user/work/trino-go-client/trino/trino_test.go:258 +0x84
  github.com/trinodb/trino-go-client/trino.(*driverStmt).exec.func1()
      /Users/user/work/trino-go-client/trino/trino.go:786 +0xc4

Goroutine 46 (running) created at:
  testing.(*T).Run()
      /usr/local/go/src/testing/testing.go:1306 +0x5b8
  testing.runTests.func1()
      /usr/local/go/src/testing/testing.go:1598 +0xac
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1259 +0x198
  testing.runTests()
      /usr/local/go/src/testing/testing.go:1596 +0x780
  testing.(*M).Run()
      /usr/local/go/src/testing/testing.go:1504 +0x928
  github.com/trinodb/trino-go-client/trino.TestMain()
      /Users/user/work/trino-go-client/trino/integration_test.go:101 +0x95c
  main.main()
      _testmain.go:129 +0x290

Goroutine 47 (finished) created at:
  github.com/trinodb/trino-go-client/trino.(*driverStmt).exec()
      /Users/user/work/trino-go-client/trino/trino.go:782 +0x614
  github.com/trinodb/trino-go-client/trino.(*driverStmt).QueryContext()
      /Users/user/work/trino-go-client/trino/trino.go:693 +0x58
  database/sql.ctxDriverStmtQuery()
      /usr/local/go/src/database/sql/ctxutil.go:82 +0x1dc
  database/sql.rowsiFromStatement()
      /usr/local/go/src/database/sql/sql.go:2794 +0x168
  database/sql.(*DB).queryDC()
      /usr/local/go/src/database/sql/sql.go:1752 +0x3e4
  database/sql.(*DB).query()
      /usr/local/go/src/database/sql/sql.go:1700 +0x124
  database/sql.(*DB).QueryContext()
      /usr/local/go/src/database/sql/sql.go:1674 +0xa4
  database/sql.(*DB).Query()
      /usr/local/go/src/database/sql/sql.go:1691 +0x4e4
  github.com/trinodb/trino-go-client/trino.TestQueryProgressWithCallbackPeriod()
      /Users/user/work/trino-go-client/trino/trino_test.go:311 +0x438
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1259 +0x198
==================
    testing.go:1152: race detected during execution of test
```

Comment (nineinchnick) [reply trino/trino_test.go:332]: You're right. I wrongly remembered how it works. No surprises here: https://go.dev/play/p/KxyNdF1TiUp

But my point is valid though, that we don't need to send any signals over the empty channel and it's enough to close it once when no more progress updates will be made.

Comment (Chaho12) [reply trino/trino_test.go:332]: >  we don't need to send any signals over the empty channel and it's enough to close it once when no more progress updates will be made.

Hmm.. I wonder why I `get race error` if I call close(qr.doneCh) at Close() and `no error` happens if I do `qr.doneCh <- struct{}{}` instead? 
Could it be because close() doc saids `has the effect of shutting down the channel after the last sent value is received` so doneCh waits for last sent value before closing?

Comment (nineinchnick) [reply trino/trino_test.go:332]: Can you push the changes where you get the race?

Comment (Chaho12) [reply trino/trino_test.go:332]: @nineinchnick here made changes

Comment (nineinchnick) [reply trino/trino_test.go:332]: Here's a full diff:
[test.txt](https://github.com/trinodb/trino-go-client/files/9382797/test.txt)

Comment (Chaho12) [reply trino/trino_test.go:332]: Regarding this code, isn't this used for channel synchronization ?  I was reluctant on using this because it will wait until statsCh has closed.
```go
 if st.statsCh != nil {
               <-st.statsCh
       }
```

Chaho12: > I previously suggested doing this in the driverRows.Close(), but it's not aware of the statsCh so it needs to be done in the driverStmt.Close()

I first thought that it would be more appropriate to close channel `AFTER rows iterator` finishes, as it also calls FETCH function. (makes sense because data is currently at FETCH status)
But, when I debugged, I noticed that `func (st *driverStmt) Close` function is called `after` `func (qr *driverRows) Close() error {`finishes, so it would further makes sense to finish when statement is done.

Chaho12: If it is allright, then I'll squash commits 
Also, I added comment to explain what driverStmt close does in [Sql.go](https://cs.opensource.google/go/go/+/refs/tags/go1.19:src/database/sql/sql.go;l=3330)
```
// Close closes statement just before releasing connection
func (st *driverStmt) Close() error {
```
- https://github.com/trinodb/trino-go-client/pull/14/commits/30d49d1238896c00c5001754d3e8b9dbab0740df

nineinchnick (approved):

losipiuk (approved):

PR #15, opened 2020-12-02
Added `Arguments` parameter to `typeSignature`

sananguliyev:
I was implementing some kind of reverse proxy for Presto but `presto-cli` was not working with my reverse proxy. During the debugging, I have noticed that the arguments parameter is missing in the type signature struct.

P.S. I have also removed `infoResponse` since it is a private struct and unused.

### DIFF
--- a/presto/presto.go
+++ b/presto/presto.go
@@ -725,9 +725,15 @@ type queryColumn struct {
 type queryData []interface{}
 
 type typeSignature struct {
-	RawType          string        `json:"rawType"`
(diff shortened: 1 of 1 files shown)


ebyhr:
@cla-bot check

losipiuk (approved):
LGTM.
Imo we should also drop `TypeArguments` and `LiteralArguments`.
Trino is not sending those back since 2019

nineinchnick:
This has been addressed in #41

PR #16, opened 2020-12-23
implement ExecContext

nineinchnick:
This PR implements `driver.ExecContext()` to be able to run queries like `USE <catalog>.<schema>` and possibly other DDL/DML queries. While this is rarely used in application code, I noticed this when trying to use the [usql](https://github.com/xo/usql) CLI client.

I also included few minor changes to satisfy `golangci-lint` linter, like adding comments and removing dead code.

When working on this I was reading https://github.com/prestosql/presto/blob/master/presto-jdbc/src/main/java/io/prestosql/jdbc/PrestoStatement.java as a reference.

### DIFF
--- a/trino/trino.go
+++ b/trino/trino.go
@@ -103,6 +103,8 @@ const (
 	trinoCatalogHeader      = "X-Presto-Catalog"
 	trinoSchemaHeader       = "X-Presto-Schema"
 	trinoSessionHeader      = "X-Presto-Session"
+	trinoSetCatalogHeader   = "X-Presto-Set-Catalog"
+	trinoSetSchemaHeader    = "X-Presto-Set-Schema"
 
 	KerberosEnabledConfig    = "KerberosEnabled"
 	kerberosKeytabPathConfig = "KerberosKeytabPath"
@@ -416,6 +418,14 @@ func (c *Conn) roundTrip(ctx context.Context, req *http.Request) (*http.Response
 			}
 			switch resp.StatusCode {
 			case http.StatusOK:
+				catalog := resp.Header.Get(trinoSetCatalogHeader)
+				if catalog != "" {
+					c.httpHeaders.Set(trinoCatalogHeader, catalog)
+				}
+				schema := resp.Header.Get(trinoSetSchemaHeader)
+				if schema != "" {
+					c.httpHeaders.Set(trinoSchemaHeader, schema)
+				}
 				return resp, nil
 			case http.StatusServiceUnavailable:
 				resp.Body.Close()
@@ -470,6 +480,7 @@ type driverStmt struct {
 var (
 	_ driver.Stmt             = &driverStmt{}
 	_ driver.StmtQueryContext = &driverStmt{}
+	_ driver.StmtExecContext  = &driverStmt{}
 )
 
 func (st *driverStmt) Close() error {
@@ -481,15 +492,38 @@ func (st *driverStmt) NumInput() int {
 }
 
 func (st *driverStmt) Exec(args []driver.Value) (driver.Result, error) {
-	return nil, ErrOperationNotSupported
+	return nil, driver.ErrSkip
+}
+
+func (st *driverStmt) ExecContext(ctx context.Context, args []driver.NamedValue) (driver.Result, error) {
+	sr, err := st.exec(ctx, args)
+	if err != nil {
+		return nil, err
+	}
+	rows := &driverRows{
+		ctx:          ctx,
+		stmt:         st,
+		nextURI:      sr.NextURI,
+		rowsAffected: sr.UpdateCount,
+	}
+	// consume all results, if there are any
+	for err == nil {
+		err = rows.fetch(true)
+	}
+	if err != nil && err != io.EOF {
+		return nil, err
+	}
+	return rows, nil
 }
 
 type stmtResponse struct {
-	ID      string    `json:"id"`
-	InfoURI string    `json:"infoUri"`
-	NextURI string    `json:"nextUri"`
-	Stats   stmtStats `json:"stats"`
-	Error   stmtError `json:"error"`
+	ID          string    `json:"id"`
+	InfoURI     string    `json:"infoUri"`
+	NextURI     string    `json:"nextUri"`
+	Stats       stmtStats `json:"stats"`
+	Error       stmtError `json:"error"`
+	UpdateType  string    `json:"updateType"`
+	UpdateCount int64     `json:"updateCount"`
 }
 
 type stmtStats struct {
@@ -553,6 +587,22 @@ func (st *driverStmt) Query(args []driver.Value) (driver.Rows, error) {
 }
 
 func (st *driverStmt) QueryContext(ctx context.Context, args []driver.NamedValue) (driver.Rows, error) {
+	sr, err := st.exec(ctx, args)
+	if err != nil {
+		return nil, err
+	}
+	rows := &driverRows{
+		ctx:     ctx,
+		stmt:    st,
+		nextURI: sr.NextURI,
+	}
+	if err = rows.fetch(false); err != nil {
+		return nil, err
+	}
+	return rows, nil
+}
+
+func (st *driverStmt) exec(ctx context.Context, args []driver.NamedValue) (*stmtResponse, error) {
 	query := st.query
 	var hs http.Header
 
@@ -588,6 +638,7 @@ func (st *driverStmt) QueryContext(ctx context.Context, args []driver.NamedValue
 	if err != nil {
 		return nil, err
 	}
+
 	defer resp.Body.Close()
 	var sr stmtResponse
 	d := json.NewDecoder(resp.Body)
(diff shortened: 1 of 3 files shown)


losipiuk:
Thanks @nineinchnick!

> I also included few minor changes to satisfy golangci-lint linter, like adding comments and removing dead code.

Can you please split PR into multiple commits so logically separate changes are in separate commits?

As for the "adding comments" part. I do not believe we want all of those. I know that some **notable** golang projects require every public member to be documented. But I strongly believe that it does not improve code quality. Rather does the opposite. We should use common sense and add comments where they are useful.

nineinchnick:
> Can you please split PR into multiple commits so logically separate changes are in separate commits?

Done!


> As for the "adding comments" part. I do not believe we want all of those. I know that some **notable** golang projects require every public member to be documented. But I strongly believe that it does not improve code quality. Rather does the opposite. We should use common sense and add comments where they are useful.

Fully agree, I tried to make them as relevant as possible.

presto/presto_test.go
```diff
@@ -275,6 +275,52 @@ func TestQueryFailure(t *testing.T) {
...
+	}
+	defer db.Close()
+
+	rows, err := db.Query(`SELECT * FROM "presto.execution:name=queryexecution"`)
+	expected := "Schema must be specified when session schema is not set"
+	if err == nil || !strings.Contains(err.Error(), expected) {
+		t.Fatalf("Expected to fail to execute query with error: %v, got: %v", expected, err)
+	}
+
+	result, err := db.Exec("USE jmx.current")
```
losipiuk, line 300: Why is `TestUnsupportedExec` below not failing with your change?

presto/presto.go
```diff
@@ -103,13 +103,17 @@ const (
 	prestoCatalogHeader     = "X-Presto-Catalog"
 	prestoSchemaHeader      = "X-Presto-Schema"
 	prestoSessionHeader     = "X-Presto-Session"
+	prestoSetCatalogHeader  = "X-Presto-Set-Catalog"
+	prestoSetSchemaHeader   = "X-Presto-Set-Schema"
 
+	// KerberosEnabledConfig is a param name
```
losipiuk, line 109: group both params in single code section and just add single comment to name a section `// Param names`

presto/presto_test.go
```diff
@@ -275,6 +275,52 @@ func TestQueryFailure(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	db, err := sql.Open("presto", dsn)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer db.Close()
+
+	rows, err := db.Query(`SELECT * FROM "presto.execution:name=queryexecution"`)
```
losipiuk, line 294: If we have access to `tpch` connector in this test I would prefer to use that instead `jmx`

presto/presto.go
```diff
@@ -1359,11 +1426,11 @@ type NullTime struct {
 
 // Scan implements the sql.Scanner interface.
 func (s *NullTime) Scan(value interface{}) error {
-	switch value.(type) {
+	switch t := value.(type) {
```
losipiuk, line 1429: use `valueType` instead of `t`. (Yeah I know golang likies single letter variables. I don't :p)

presto/presto.go
```diff
@@ -416,6 +420,14 @@ func (c *Conn) roundTrip(ctx context.Context, req *http.Request) (*http.Response
 			}
 			switch resp.StatusCode {
 			case http.StatusOK:
+				catalog := resp.Header.Get(prestoSetCatalogHeader)
+				if catalog != "" {
+					c.httpHeaders.Set(prestoCatalogHeader, catalog)
+				}
+				schema := resp.Header.Get(prestoSetSchemaHeader)
```
losipiuk, line 427: should we handle `X-Set-Path` here too?

losipiuk (commented):
Thanks. Mostly looks good. Some cleanup comments.

And also please capitalize commit messages.

presto/presto_test.go
nineinchnick: I totally missed that! It looks like that test was invalid as it only checked if the err is nil. I removed this one and fixed the following one for unsupported trx by adding an assert for a specific error message.

presto/presto.go
nineinchnick: All exported vars and functions need to have a comment that begins with their name and reads like a proper english sentence. These are an example where the name is pretty much self explanatory, so the comment doesn't add much. We have three options here:
* change the comment to one that disables the linter rule on this line
* disable this specific linter for the whole repo
* leave these comments as is

WDYT?

presto/presto_test.go
nineinchnick: Done!

presto/presto.go
nineinchnick: Sorry but I need to push back here. The var we instantiate here is not the type but of specific type so `typedValue` would be slightly better. But in both cases it's actually a time so a more specific name is even better. But `time` is the name of the package, hence `t`. This is idiomatic, `t` is used a lot in the examples from stdlib: https://golang.org/pkg/time/

presto/presto.go
nineinchnick: I'm not sure. It's not set in the initial connection. In what cases the server would send it and expect it to be set in future requests?

nineinchnick: LMK if you think we should add this for consistency.

presto/presto.go
losipiuk: I would go for `disable this specific linter for the whole repo`. This one specifically makes more mess than benefit. Instead IMO we should use review-process to ensure that comments are present where they actually improve code readability (whether it is exported function, private one, or some code snippet within a function)

presto/presto.go
losipiuk: fair :) Let's leave t.

presto/presto.go
losipiuk: Actually there are more of those:

https://github.com/trinodb/trino/blob/83bcdffc4c8dd0ce377c5f12609c69b9088ba6a0/presto-client/src/main/java/io/prestosql/client/PrestoHeaders.java#L27-L32

I think we do not have to support all of those initially. But we should at least notice that they are present in the response and return an error to the caller, if we get an unsupported `X-Prest-Set-...` header. Otherwise we will pretend that we support `SET SESSION ...` syntax, but we would not send set session properties in followup queries.

Could you please add validation and test coverage?

presto/presto.go
nineinchnick: I reverted these changes. It turned out that by default `golangci-lint` ignores these specific errors from the `golint` linter but my editor overrides this. The full list is here: https://golangci-lint.run/usage/configuration/

Since we don't have any linters in the pipeline I'm not going to add any configs here. This is out of scope of this PR.

presto/presto.go
nineinchnick: Done! Added this as a separate commit.

trino/trino.go
```diff
@@ -114,6 +121,19 @@ const (
 	SSLCertPathConfig        = "SSLCertPath"
 )
 
+var (
+	headersMap = map[string]string{
```
losipiuk, line 125: nit: rename to `responseToRequestHeaderMap`?

trino/trino.go
```diff
@@ -114,6 +121,19 @@ const (
...
+		trinoSetSchemaHeader:  trinoSchemaHeader,
+		trinoSetCatalogHeader: trinoCatalogHeader,
+	}
+	unsupportedHeaders = []string{
```
losipiuk, line 129: nit: rename to `unsupportedResponseHeaders`

trino/trino_test.go
```diff
@@ -321,6 +321,23 @@ func TestExec(t *testing.T) {
 	}
 }
 
+func TestUnsupportedHeader(t *testing.T) {
```
losipiuk, line 324: Can you also add and integration test calling out to 
* `SET SESSION grouped_execution=true`
* `SET ROLE dummy`
* `SET PATH dummy`
* `RESET SESSION grouped_execution` (not sure if this one will work; may not send X-Presto-Clear-Session header if it was not set before, which we cannot do)

trino/trino_test.go
```diff
@@ -275,6 +275,52 @@ func TestQueryFailure(t *testing.T) {
 	}
 }
 
+func TestExec(t *testing.T) {
```
losipiuk, line 278: I think this should go to `integration_test.go`

trino/trino_test.go
```diff
@@ -275,6 +275,52 @@ func TestQueryFailure(t *testing.T) {
...
+	dsn, err := c.FormatDSN()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	db, err := sql.Open("trino", dsn)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer db.Close()
```
losipiuk, line 292: after moving to `integration_test.go` you should be able to replace this with
```
	db := integrationOpen(t)
	defer db.Close()
```

losipiuk (commented):
Awesome. Minor comments.

trino/trino.go
nineinchnick: Done!

trino/trino.go
nineinchnick: Done!

trino/trino_test.go
nineinchnick: Done! Actually only `SET PATH` failed since there's some client capability negotiation mechanism, this client doesn't report any capabilities and setting PATH is not allowed by default.

trino/trino_test.go
nineinchnick: Right, there's a few other tests in here that should be moved but I'll leave this to another PR.

trino/trino_test.go
nineinchnick: Done!

losipiuk:
@cla-bot check

losipiuk:
@cla-bot check

nineinchnick:
@cla-bot check

nineinchnick:
@cla-bot check

losipiuk (approved):

losipiuk:
Merged, thanks!

PR #17, opened 2020-12-23
Delete cancelled query

nineinchnick:
When using `QueryContext()` and canceling the context the driver should send a DELETE request to the server to stop the query from being executed and wasting resources.

I also added a 5s timeout to `go test` in `run.sh` and my editor ran `shfmt` on it, I hope it's not a problem. That fixed a few `shellcheck` issues as well (quoting, subshells etc).

I also made type comparison case insensitive, since the server might be using upper case for some types, like `INTERVAL`.

### DIFF
--- a/trino/trino.go
+++ b/trino/trino.go
@@ -910,7 +910,7 @@ func newTypeConverter(typeName string) *typeConverter {
 // parses presto types, e.g. array(varchar(10)) to "array", "varchar"
 // TODO: Use queryColumn.TypeSignature instead.
 func parseType(name string) []string {
-	parts := strings.Split(name, "(")
+	parts := strings.Split(strings.ToLower(name), "(")
 	if len(parts) == 1 {
 		return parts
 	}
--- a/trino/trino.go
+++ b/trino/trino.go
@@ -432,6 +432,7 @@ func (c *Conn) roundTrip(ctx context.Context, req *http.Request) (*http.Response
 			}
 			client := c.httpClient
 			client.Timeout = timeout
+			req.Cancel = ctx.Done()
 			resp, err := client.Do(req)
 			if err != nil {
 				return nil, &ErrQueryFailed{Reason: err}
@@ -525,6 +526,7 @@ func (st *driverStmt) ExecContext(ctx context.Context, args []driver.NamedValue)
 	rows := &driverRows{
 		ctx:          ctx,
 		stmt:         st,
+		queryID:      sr.ID,
 		nextURI:      sr.NextURI,
 		rowsAffected: sr.UpdateCount,
 	}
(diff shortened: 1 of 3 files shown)


presto/presto.go
```diff
@@ -626,29 +627,27 @@ type driverRows struct {
...
-		defer cancel()
-		resp, err := qr.stmt.conn.roundTrip(ctx, req)
-		if err != nil {
-			qferr, ok := err.(*ErrQueryFailed)
-			if ok && qferr.StatusCode == http.StatusNoContent {
-				qr.nextURI = ""
-				return nil
-			}
-			return err
+	if qr.nextURI == "" {
```
losipiuk, line 630: note for future PRs. The changes to this method are just refactorings. Woudl be nice to have it as a separate preparatory commit.

presto/presto_test.go
```diff
@@ -191,6 +191,79 @@ func TestRoundTripCancellation(t *testing.T) {
 	}
 }
 
+func TestQueryContextCancellation(t *testing.T) {
```
losipiuk, line 194: Can we test it as an integration test and run queries against actual Trino instance run in docker?. 
Testing against mock HTTP server works but is very verbose and simulates specific server behavior some of which may be treated as an implementation detail.

We should be able to schedule a real long running query (e.g. `SELECT COUNT(*) from tpch.sf10.lineitem`. Cancel it from sub routine. And then verify with another query that original query was cancelled on the server side by using `system.runtime.queries` table (while query is still running you can determine its query id by querying same table).

presto/presto.go
```diff
@@ -832,7 +832,7 @@ func parseType(name string) []string {
 
 // ConvertValue implements the driver.ValueConverter interface.
 func (c *typeConverter) ConvertValue(v interface{}) (driver.Value, error) {
-	switch c.parsedType[0] {
+	switch strings.ToLower(c.parsedType[0]) {
```
losipiuk, line 835: Please move lowercasing to `parseType` method

presto/presto.go
nineinchnick: Duly noted

presto/presto_test.go
nineinchnick: I replaced this test with an integration one. While writing the test I found multiple issues so please review the whole PR again.

presto/presto.go
nineinchnick: Done!

trino/trino.go
```diff
@@ -836,6 +836,14 @@ func parseType(name string) []string {
 	return parts
 }
 
+func mapStrings(vs []string, f func(string) string) []string {
```
losipiuk, line 839: I think it would be more easier to just add 
```
	name = strings.ToLower(name)
```
as a first line in `parseType` :)

losipiuk (commented):
Look decent thanks :)

trino/trino.go
nineinchnick: 🤦 I guess that's exactly why we have code reviews! Done, thanks!

nineinchnick:
@cla-bot check

nineinchnick:
@cla-bot check

losipiuk:
@nineinchnick rebase needed

nineinchnick:
Rebased and ready to go!

losipiuk:
Merged, thanks!

PR #18, opened 2020-12-29
Rebrand to Trino

losipiuk:


trino/trino.go
```diff
@@ -78,31 +78,31 @@ import (
...
 const (
 	preparedStatementHeader = "X-Presto-Prepared-Statement"
-	preparedStatementName   = "_presto_go"
-	prestoUserHeader        = "X-Presto-User"
-	prestoSourceHeader      = "X-Presto-Source"
-	prestoCatalogHeader     = "X-Presto-Catalog"
-	prestoSchemaHeader      = "X-Presto-Schema"
-	prestoSessionHeader     = "X-Presto-Session"
+	preparedStatementName   = "_trino_go"
+	trinoUserHeader         = "X-Presto-User"
```
nineinchnick, line 101: How about removing the prefix so it's just `userHeader`? Since the package name is already `trino`, the prefix stutters and there's no other `userHeader` in the whole package.

trino/trino.go
```diff
@@ -122,7 +122,7 @@ var _ driver.Driver = &sqldriver{}
 
 // Config is a configuration that can be encoded to a DSN string.
 type Config struct {
-	PrestoURI          string            // URI of the Presto server, e.g. http://user@localhost:8080
+	TrinoURI           string            // URI of the Trino server, e.g. http://user@localhost:8080
```
nineinchnick, line 125: Same here, how about just `URI`?

trino/trino.go
```diff
@@ -183,11 +183,11 @@ func (c *Config) FormatDSN() (string, error) {
...
 		}
 	}
-	prestoURL.RawQuery = query.Encode()
-	return prestoURL.String(), nil
+	trinoURL.RawQuery = query.Encode()
+	return trinoURL.String(), nil
 }
 
-// Conn is a presto connection.
+// Conn is a trino connection.
```
nineinchnick, line 190: nitpick: this should be capitalized, so `Trino`

trino/trino.go
```diff
@@ -203,47 +203,47 @@ var (
...
 func newConn(dsn string) (*Conn, error) {
-	prestoURL, err := url.Parse(dsn)
+	trinoURL, err := url.Parse(dsn)
 	if err != nil {
-		return nil, fmt.Errorf("presto: malformed dsn: %v", err)
+		return nil, fmt.Errorf("trino: malformed dsn: %v", err)
 	}
 
-	prestoQuery := prestoURL.Query()
+	trinoQuery := trinoURL.Query()
```
nineinchnick, line 211: Same here, how about just `query`?

trino/trino_test.go
```diff
@@ -330,83 +330,83 @@ func TestTypeConversion(t *testing.T) {
 		t.Fatal(err)
 	}
 	testcases := []struct {
-		PrestoType                       string
-		PrestoResponseUnmarshalledSample interface{}
-		ExpectedGoValue                  interface{}
+		TrinoType                       string
```
nineinchnick, line 333: How about `s/TrinoType/dataType`? These doesn't have to be exported and the prefix is unnecessary

nineinchnick (approved):
I was about to ask if the readme should mention that this driver could also work with PrestoDB but I guess since integration tests only cover Trino this answers my question. I like that having a more distinct name will clear up the confusion between this repo and [prestodb](https://github.com/prestodb/presto-go-client).

trino/integration_tls_test.go
```diff
@@ -75,31 +75,31 @@ func testSimpleQuery(t *testing.T, dsn string) {
 // newTLSReverseProxy creates a TLS integration test server.
 func newTLSReverseProxy(t *testing.T) *httptest.Server {
 	dsn := integrationServerDSN(t)
-	prestoURL, _ := url.Parse(dsn)
+	trinoURL, _ := url.Parse(dsn)
```
findepi, line 78: serverURL

trino/trino.go
```diff
@@ -138,7 +138,7 @@ type Config struct {
 
 // FormatDSN returns a DSN string from the configuration.
 func (c *Config) FormatDSN() (string, error) {
-	prestoURL, err := url.Parse(c.PrestoURI)
+	trinoURL, err := url.Parse(c.TrinoURI)
```
findepi, line 141: serverURL?

trino/trino.go
```diff
@@ -203,47 +203,47 @@ var (
 )
 
 func newConn(dsn string) (*Conn, error) {
-	prestoURL, err := url.Parse(dsn)
+	trinoURL, err := url.Parse(dsn)
```
findepi, line 206: serverURL

README.md
```diff
@@ -1,9 +1,9 @@
-# Presto client
+# Trino client
 
-A [Presto](https://prestosql.io) client for the [Go](https://golang.org) programming language.
+A [Trino](https://trinodb.io) client for the [Go](https://golang.org) programming language.
```
findepi, line 3: trinodb.io -> trino.io

(same below too)

README.md
```diff
@@ -61,21 +61,21 @@ Both HTTP Basic and Kerberos authentication are supported.
 
 #### HTTP Basic authentication
 
-If the DSN contains a password, the client enables HTTP Basic authentication by setting the `Authorization` header in every request to presto.
+If the DSN contains a password, the client enables HTTP Basic authentication by setting the `Authorization` header in every request to trino.
```
findepi, line 64: trino -> Trino

same in few other places in README

CONTRIBUTING.md
```diff
@@ -1,9 +1,9 @@
-# Contributing to Presto
+# Contributing to Trino
 
 ## Contributor License Agreement ("CLA")
 
-In order to accept your pull request, we need you to [submit a CLA](https://github.com/prestosql/cla).
+In order to accept your pull request, we need you to [submit a CLA](https://github.com/triondb/cla).
```
findepi, line 5: triondb -> trinodb

findepi (approved):
Some typos. Further renames are optional IMO and can be viewed as out of scope, if you want.

trino/trino.go
losipiuk: renamed to `serverURI`

trino/trino.go
losipiuk: It is subjective - but I will keep prefix here. This is constant definition, and having prefix here points out that is constant specific to Trino protocol. Not some generic HTTP header which we happen to use with Trino.

PR #19, opened 2021-01-16
Finish rename to Trino

nineinchnick:
Finish rename to Trino, notably use `X-Trino` in HTTP headers.


nineinchnick:
@losipiuk can you create a (any) version tag after merging this?

losipiuk (approved):

losipiuk:
> @losipiuk can you create a (any) version tag after merging this?

I tagged as `0.300.0`. We can start versioning from this. cc: @findepi

Issue #25, opened 2021-11-10
Support for more than one header?


a-urth:
Hello!

We at Lyft would benefit from possibility to pass more than one header in order to correctly handle requests through our custom proxy. But unfortunatelly currently trino client supports only one header - `X-Trino-User` (https://github.com/trinodb/trino-go-client/blob/master/trino/trino.go#L652).

What do You think about adding support for multiple headers? Lets say any header prefixed with `X-Trino`?

If that sounds ok, then I can prepare a PR for that.

Cheers

hashhar:
Seems similar to https://github.com/trinodb/trino-python-client/pull/121 for the Python client.

It makes sense to me. cc: @losipiuk ?

nineinchnick:
Should we close this issue now or are we waiting for a release?

hashhar:
Fixed in https://github.com/trinodb/trino-go-client/pull/31. Will be part of next release.

PR #26, opened 2021-11-12
Send all the headers which are correctly prefixed within request

a-urth:
https://github.com/trinodb/trino-go-client/issues/25

### DIFF
--- a/trino/trino.go
+++ b/trino/trino.go
@@ -101,18 +101,19 @@ var (
 const (
 	preparedStatementHeader    = "X-Trino-Prepared-Statement"
 	preparedStatementName      = "_trino_go"
-	trinoUserHeader            = "X-Trino-User"
-	trinoSourceHeader          = "X-Trino-Source"
-	trinoCatalogHeader         = "X-Trino-Catalog"
-	trinoSchemaHeader          = "X-Trino-Schema"
-	trinoSessionHeader         = "X-Trino-Session"
-	trinoSetCatalogHeader      = "X-Trino-Set-Catalog"
-	trinoSetSchemaHeader       = "X-Trino-Set-Schema"
-	trinoSetPathHeader         = "X-Trino-Set-Path"
-	trinoSetSessionHeader      = "X-Trino-Set-Session"
-	trinoClearSessionHeader    = "X-Trino-Clear-Session"
-	trinoSetRoleHeader         = "X-Trino-Set-Role"
-	trinoExtraCredentialHeader = "X-Trino-Extra-Credential"
+	trinoHeaderPrefix = `X-Trino-`
+	trinoUserHeader            = trinoHeaderPrefix+`User`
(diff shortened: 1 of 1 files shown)


nineinchnick (approved):

a-urth:
@cla-bot check

a-urth:
There are two tests failing currently:
- `TestIntegrationUnsupportedHeader` - this one I'm not sure why its happening, I'll appreciate if someone would explain what could be the problem
- `TestExtraCredentials` - this one fails due to how query parameters are attached to URL - because they are taken from the map, order is not guaranteed and comparison is done with hardcoded string. This test and all other tests which make comparison same way should be fixed so I'll do it in separate PR

a-urth:
@cla-bot check

trino/trino.go
```diff
@@ -101,18 +101,19 @@ var (
 const (
 	preparedStatementHeader    = "X-Trino-Prepared-Statement"
```
hashhar, line 102: This one can also use the `trinoHeaderPrefix`.

hashhar (commented):
Please submit a PR to fix tests to not depend on header order (or a separate commit in this PR itself - make it the first commit) so that we can merge it first and then the tests pass on this one so that this one can be merged too.

trino/trino.go
```diff
@@ -101,18 +101,19 @@ var (
...
+	trinoCatalogHeader         = trinoHeaderPrefix+`Catalog`
+	trinoSchemaHeader          = trinoHeaderPrefix+`Schema`
+	trinoSessionHeader         = trinoHeaderPrefix+`Session`
+	trinoSetCatalogHeader      = trinoHeaderPrefix+`Set-Catalog`
+	trinoSetSchemaHeader       = trinoHeaderPrefix+`Set-Schema`
+	trinoSetPathHeader         = trinoHeaderPrefix+`Set-Path`
+	trinoSetSessionHeader      = trinoHeaderPrefix+`Set-Session`
+	trinoClearSessionHeader    = trinoHeaderPrefix+`Clear-Session`
+	trinoSetRoleHeader         = trinoHeaderPrefix+`Set-Role`
+	trinoExtraCredentialHeader = trinoHeaderPrefix+`Extra-Credential`
```
ebyhr, line 116: These change reminds me https://github.com/trinodb/trino-python-client/pull/28. I prefer the original values.

a-urth: sure! i don't like how it looks either, although i made this change because if header prefix and headers themselves are separate variables then there is a risk that only one will be changed.

a-urth:
Hey guys, I fixed failing unit test within [this PR](https://github.com/trinodb/trino-go-client/pull/27) and also I decided to cleanup unit tests in general. Make them more idiomatic and verbose.

a-urth:
closing in favour of #31

Issue #11, opened 2020-08-24
Migrate test assertions to github.com/stretchr/testify/assert


losipiuk:

hashhar:
Fixed by #27

PR #27, opened 2021-11-18
Unit tests cleanup

a-urth:
There are two major things within this PR
- ordering for string items in `FormatDSN` - it doesn't affect any existing functionality although it makes tests based on URL comparison much easier since items' order within variadic query args will be consistent
- unit tests cleanup - use `testify` library to make assertions shorter and more verbose, use `t.Cleanup` instead of `defer`

### DIFF
--- a/trino/trino_test.go
+++ b/trino/trino_test.go
@@ -19,10 +19,10 @@ import (
 	"database/sql"
 	"encoding/json"
 	"fmt"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"net/http"
 	"net/http/httptest"
-	"reflect"
-	"strings"
 	"testing"
 	"time"
 )
@@ -32,14 +32,13 @@ func TestConfig(t *testing.T) {
 		ServerURI:         "http://foobar@localhost:8080",
 		SessionProperties: map[string]string{"query_priority": "1"},
 	}
+
 	dsn, err := c.FormatDSN()
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err)
+
 	want := "http://foobar@localhost:8080?session_properties=query_priority%3D1&source=trino-go-client"
-	if dsn != want {
-		t.Fatal("unexpected dsn:", dsn)
-	}
+
+	assert.Equal(t, want, dsn)
 }
 
 func TestConfigSSLCertPath(t *testing.T) {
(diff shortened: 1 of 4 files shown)


a-urth:
Integration test if failing, thats the same one as in my other PR. I have a feeling that behaviour might have change in trinodb used for integration tests and was not reflected in test itself.

hashhar:
Indeed, master fails on latest image too - https://github.com/trinodb/trino-go-client/pull/28.

hashhar:
Closing and re-opening to get CI to re-run.

hashhar (commented):
Revert the dependency changes if not required (since it seems it just replaces stdlib tests with testify?)

Looks good otherwise.

a-urth:
Reverted version bump for testify lib, since it was already used in the project although not directly.

@hashhar as for tests, this lib is very useful, since you don't need to check for error (or other condition) and then fail test. Tests functionally the same, but just look way cleaner

trino/trino_test.go
```diff
@@ -418,28 +412,27 @@ func TestTypeConversion(t *testing.T) {
...
-			if err != nil {
-				t.Fatal(err)
-			}
-			if !reflect.DeepEqual(v, tc.ExpectedGoValue) {
-				t.Fatalf("unexpected data from sample:\nhave %+v\nwant %+v", v, tc.ExpectedGoValue)
-			}
+			require.NoError(t, err)
+
+			require.Truef(t,
+				reflect.DeepEqual(v, tc.ExpectedGoValue),
```
nineinchnick, line 434: Could you use `require.Equal()` here too? It seems to fall back to `DeepEqual` anyway: https://github.com/stretchr/testify/blob/master/assert/assertions.go#L65

nineinchnick (commented):
I don't mind the explicit conditions in tests, but I think testify would produce better error messages, adding unexpected errors and expected vs actual values.

But if we were to switch to testify, we should be consistent, and use it in all tests, including integration tests.

nineinchnick:
This PR addresses #11

hashhar:
Agreed, let's extract a commit / PR though which adds the `sort` to ensure tests always get predictable ordering.

And keep the testify change in a separate commit / PR.

trino/trino_test.go
a-urth: Sure thing, will do.

hashhar:
Splitting into commits is also fine, doesn't have to be a separate PR. It's very difficult to notice in future that something functional changed (like sorting values) in a large diff where it looks like the only change is moving from stdlib tests to testify.

a-urth:
@hashhar I've changed PR to have 2 main commits (and 1 additional from review):
- one with change to order variadic query args
- another one with tests overhaul

hashhar (approved):
Thanks. Changes look good to me.

Can you squash the last 2 commits into one (so that in the end there are only 2 commits)?

I'll merge once CI finishes.

a-urth:
@hashhar i've joined last two commits with unit tests' restructure and review comments into one

hashhar:
@a-urth Thanks. Looks good. Can we add some detail to second commit message like `Migrate test assertions to github.com/stretchr/testify/assert`?

hashhar:
See https://github.com/trinodb/trino/blob/master/.github/DEVELOPMENT.md#git-merge-strategy

hashhar (approved):

hashhar:
Merged. Thanks for the improvements. 🎉

PR #29, opened 2021-11-19
Fix broken role integration test

ebyhr:
Version 362 started disallowing to set roles which don't exist.


hashhar (commented):
I'll prepare equivalent of https://github.com/trinodb/trino-python-client/pull/128 for this repo too so that we can validate this change via CI.

I think you might need a conditional in the test to make it pass on both versions?
Or do we want to switch to using a newer image for testing?

a-urth (approved):

hashhar (approved):

PR #35, opened 2022-01-05
Simplify prepared statement parameters handling

a-urth:
Problem
If my understanding of the logic below is correct then named arguments which are not correctly prefixed (X-Trino-...) are considered as prepared statement's parameters. If so then logic was a bit convoluted - for example there was no need to check for empty prepared statement header for each argument.

Solution
Restructure argument parsing logic so headers and non headers (prepared statement parameters) cases are more explicit and verbose. Add corresponding comments and improve readability.


nineinchnick:
@a-urth sorry for not responding for over 2 years, but can you rebase it?

PR #41, opened 2022-06-12
Return scan types in rows.ColumnTypes()

nineinchnick:
This PR implements the required interfaces so that rows' `ColumnTypes()` include scan types. It also adds length, precision,  and size to applicable data types.

[The docs for database type name](https://pkg.go.dev/database/sql/driver#RowsColumnTypeDatabaseTypeName) say it should be returned in uppercase, without any length or precision. This would be a breaking change, since before it was returned lowercase. I'm also not entirely sure how to handle composite types, so I return them as-is, ex. `ROW(VARCHAR(10), BIGINT)`.

I used the type signatures from the response instead of trying to parse the text.

My use case for this is I need to know the data types before trying to read any rows. Otherwise, I'd have to add a lot of boilerplate code to hold the first row and process it later.

### DIFF
--- a/trino/trino.go
+++ b/trino/trino.go
@@ -66,7 +66,6 @@ import (
 	"net/http"
 	"net/url"
 	"reflect"
-	"regexp"
 	"sort"
 	"strconv"
 	"strings"
@@ -98,6 +97,9 @@ var (
 
 	// ErrUnsupportedHeader indicates that the server response contains an unsupported header.
 	ErrUnsupportedHeader = errors.New("trino: server response contains an unsupported header")
+
+	// ErrInvalidResponseType indicates that the server returned an invalid type definition.
+	ErrInvalidResponseType = errors.New("trino: server response contains an invalid type")
 )
 
 const (
@@ -752,6 +754,10 @@ type driverRows struct {
 
 var _ driver.Rows = &driverRows{}
 var _ driver.Result = &driverRows{}
+var _ driver.RowsColumnTypeScanType = &driverRows{}
+var _ driver.RowsColumnTypeDatabaseTypeName = &driverRows{}
+var _ driver.RowsColumnTypeLength = &driverRows{}
+var _ driver.RowsColumnTypePrecisionScale = &driverRows{}
 
 // Close closes the rows iterator.
 func (qr *driverRows) Close() error {
@@ -796,14 +802,24 @@ func (qr *driverRows) Columns() []string {
 	return qr.columns
 }
 
-var coltypeLengthSuffix = regexp.MustCompile(`\(\d+\)$`)
-
 func (qr *driverRows) ColumnTypeDatabaseTypeName(index int) string {
-	name := qr.coltype[index].typeName
-	if m := coltypeLengthSuffix.FindStringSubmatch(name); m != nil {
-		name = name[0 : len(name)-len(m[0])]
+	typeName := qr.coltype[index].parsedType[0]
+	if typeName == "map" || typeName == "array" || typeName == "row" {
+		typeName = qr.coltype[index].typeName
 	}
-	return name
+	return strings.ToUpper(typeName)
+}
+
+func (qr *driverRows) ColumnTypeScanType(index int) reflect.Type {
+	return qr.coltype[index].scanType
+}
+
+func (qr *driverRows) ColumnTypeLength(index int) (int64, bool) {
+	return qr.coltype[index].size.value, qr.coltype[index].size.hasValue
+}
+
+func (qr *driverRows) ColumnTypePrecisionScale(index int) (precision, scale int64, ok bool) {
+	return qr.coltype[index].precision.value, qr.coltype[index].scale.value, qr.coltype[index].precision.hasValue
 }
 
 // Next is called to populate the next row of data into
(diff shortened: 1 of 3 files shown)


nineinchnick:
This supersedes #15

nineinchnick:
I rebased this locally and some tests failed, I'm working on it.

trino/trino.go
```diff
@@ -865,10 +879,26 @@ type queryColumn struct {
...
 type typeSignature struct {
-	RawType          string        `json:"rawType"`
-	TypeArguments    []interface{} `json:"typeArguments"`
-	LiteralArguments []interface{} `json:"literalArguments"`
+	RawType   string         `json:"rawType"`
+	Arguments []typeArgument `json:"arguments"`
+}
+
+type typeArgument struct {
+	// Kind is one of: TYPE, NAMED_TYPE, LONG, VARIABLE
```
losipiuk, line 893: nit: you could have a type for that.

nineinchnick: done!

trino/trino.go
```diff
@@ -932,48 +971,185 @@ func (qr *driverRows) fetch(allowEOF bool) error {
...
-	typeName   string
-	parsedType []string // e.g. array, array, varchar, for [][]string
+	typeName      string
+	parsedType    []string
+	scanType      reflect.Type
+	typeSignature typeSignature
+	hasPrecision  bool
+	precision     int64
+	scale         int64
+	hasSize       bool
```
losipiuk, line 1031: nit: is there sth like OptionalInt64? Would be nicer. Also applies to precision and scale.

trino/trino.go
```diff
@@ -932,48 +971,185 @@ func (qr *driverRows) fetch(allowEOF bool) error {
...
+func newTypeConverter(typeName string, signature typeSignature) *typeConverter {
+	result := &typeConverter{
+		typeName:      typeName,
+		parsedType:    getNestedTypes([]string{}, signature),
+		typeSignature: signature,
+	}
+	result.scanType = getScanType(result.parsedType)
+	switch signature.RawType {
+	case "char", "varchar":
+		if len(signature.Arguments) > 0 && signature.Arguments[0].Kind == "LONG" {
```
losipiuk, line 1044: can `signature.Arguments[0].Kind` be sth different that `LONG`? Should we just assert here that it is?

trino/trino.go
```diff
@@ -932,48 +971,185 @@ func (qr *driverRows) fetch(allowEOF bool) error {
...
+			result.hasSize = true
+			result.size = signature.Arguments[0].long
+		}
+	case "decimal":
+		if len(signature.Arguments) > 1 {
+			result.hasPrecision = true
+			if signature.Arguments[0].Kind == "LONG" {
```
losipiuk, line 1051: same here - feels like type of args should be just asserted on.

trino/trino.go
```diff
@@ -932,48 +971,185 @@ func (qr *driverRows) fetch(allowEOF bool) error {
...
+	return result
 }
 
-func newTypeConverter(typeName string) *typeConverter {
-	return &typeConverter{
-		typeName:   typeName,
-		parsedType: parseType(typeName),
+func getNestedTypes(types []string, signature typeSignature) []string {
+	types = append(types, signature.RawType)
+	if len(signature.Arguments) > 0 {
```
losipiuk, line 1064: what about the case when there are more than single type argument?
You just look at the first one. It feels to me that the `if` should be `len(signature.Arguments) == 0`, or we should explicitly support only specific types here.

trino/trino.go
```diff
@@ -932,48 +971,185 @@ func (qr *driverRows) fetch(allowEOF bool) error {
...
+	case "bigint":
+		v = sql.NullInt64{}
+	case "real", "double":
+		v = sql.NullFloat64{}
+	case "date", "time", "time with time zone", "timestamp", "timestamp with time zone":
+		v = sql.NullTime{}
+	case "map":
+		v = NullMap{}
+	case "array":
+		if len(typeNames) <= 1 {
```
losipiuk, line 1095: what exactly does this mean?

trino/trino.go
```diff
@@ -932,48 +971,185 @@ func (qr *driverRows) fetch(allowEOF bool) error {
...
 	}
 }
 
 type typeConverter struct {
-	typeName   string
-	parsedType []string // e.g. array, array, varchar, for [][]string
+	typeName      string
+	parsedType    []string
+	scanType      reflect.Type
+	typeSignature typeSignature
```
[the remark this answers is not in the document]
nineinchnick: We're not. This is an incremental improvement that replaces parsing a text type (splitting on `(`) with reading the nested types from the response, but it doesn't properly handle all nested types yet - you did notice all these zero indexes in other places. Anyway, I'll remove this from here now.

trino/trino.go
nineinchnick: This is to avoid reading unset array indices. I'll change this into an assert.

trino/trino.go
nineinchnick: This is an incremental improvement only to get the types from the response instead of parsing the text type and splitting on `(`. Can we address this in a follow-up PR?

trino/trino.go
nineinchnick: There's nothing built-in but I added a struct like that.

trino/trino.go
nineinchnick: I added an assert.

trino/trino.go
```diff
@@ -932,48 +971,185 @@ func (qr *driverRows) fetch(allowEOF bool) error {
...
+	}
+	result.scanType = getScanType(result.parsedType)
+	switch signature.RawType {
+	case "char", "varchar":
+		if len(signature.Arguments) > 0 && signature.Arguments[0].Kind == "LONG" {
+			result.hasSize = true
+			result.size = signature.Arguments[0].long
+		}
+	case "decimal":
+		if len(signature.Arguments) > 1 {
```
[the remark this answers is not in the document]
nineinchnick: I separated these checks, so even if the server returns only the precision, that's how it'll show up in the scan type.

trino/trino.go
nineinchnick: Added an assert

trino/trino.go
nineinchnick: BTW there would be multiple arguments only for composite types like row and we wouldn't return scan types for them anyway, leaving scanning completely to the user.

nineinchnick: Ah, I read this again and you're right, it shouldn't only grab the first arg when there are multiple ones. I changed this to `== 1`.

losipiuk (approved):

PR #44, opened 2022-06-12
Support SET/RESET SESSION and PREPARE

nineinchnick:
Handle the following response headers:
* X-Trino-Set-Session
* X-Trino-Clear-Session
* X-Trino-Added-Prepare
* X-Trino-Deallocated-Prepare


losipiuk (approved):
please rebase (conflicts).

Btw. I hate those golang's short meaningless variable names `v`, `v2` :) But I guess this is convention we should follow nevertheless.

losipiuk:
rebased

PR #45, opened 2022-07-08
Migrate integration test setup to go

nineinchnick:
Setup the container for integration tests in Go instead of bash. This has the following advantages:
* all tests can be executed using the regular `go test` command, contributors don't have to know about any extra commands
* can use a random port and avoid conflicts with any other Trino containers running
* integration tests can be skipped using the built-in `-short` flag

I added the `-no_cleanup` flag to skip container setup when running integration tests repeatedly.

Supersedes #43 because I had to update the dependencies anyway.


losipiuk:
Why those renames `integration_tests/...` to `trino/....`?

trino/integration_test.go
```diff
@@ -19,46 +19,103 @@ import (
...
-}
 
-// integrationServerDSN returns the URL of the integration test server.
-func integrationServerDSN(t *testing.T) string {
-	if dsn := *integrationServerFlag; dsn != "" {
-		return dsn
+	var err error
+	if *integrationServerFlag == "" && !testing.Short() {
+		pool, err = dt.NewPool("")
+		pool.MaxWait = 5 * time.Minute
```
losipiuk, line 61: move below `err` check?

Also does it correspond to 20s timeout for trino being spinned up from the bash version? If so maybe make it a bit shorter wdyt?

nineinchnick: Fixed, thanks! I lowered this to 1 minute. In the bash script, the timeout didn't account for the `trino` command to be blocking. Here it's tracked independently.

nineinchnick:
I moved properties files into the `trino` package because this is where they're needed and where test setup and tests are defined.

PR #46, opened 2022-07-09
Support ROW types

### LINKED ISSUE #39: Row data type not supported
Trino supports row data type as mentioned in the official documentation.

But there is no support for the data parsing in case we have some data in `Row` format

I had a look at `ConvertValue` function which does not have a switch case for the data type.
As it is called internally from `nextLocked` in sql.go while performing `Next()` to iterate and extract data, it is not possible to have a column as Row type, unless we manually cast all query to json.

nineinchnick:
`ROW` columns are just slices of values of different types.

Fixes #39

### DIFF
--- a/trino/trino_test.go
+++ b/trino/trino_test.go
@@ -19,12 +19,13 @@ import (
 	"database/sql"
 	"encoding/json"
 	"fmt"
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
 	"net/http"
 	"net/http/httptest"
 	"testing"
 	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestConfig(t *testing.T) {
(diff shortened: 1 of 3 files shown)


trino/trino_test.go
```diff
@@ -410,6 +411,22 @@ func TestTypeConversion(t *testing.T) {
 			ResponseUnmarshalledSample: nil,
 			ExpectedGoValue:            nil,
 		},
+		{
+			// rows return data as-is for slice scanners
```
losipiuk, line 415: Is that ok? What is the alternative?

losipiuk (approved):

trino/trino_test.go
nineinchnick: This is ok, it's similar to how arrays are handled, but in rows, values can be of different types. I'm not sure it's possible to construct an anonymous struct using reflection, so this is as good as we can get. Users would can consume such a slice directly, doing type assertion on values, but they should expect which field has which type. Or they can implement the scanner interface in a struct, which would populate the struct. This would be more idiomatic and this is what I documented in the README.

PR #48, opened 2022-07-10
Support passing slices as query args

### LINKED ISSUE #20: Unable to use []string as query parameter
From what i see in `serial.go`, there should be support for slice of strings as parameters.

But when I try to execute something like
```golang
s.db.QueryContext(ctx, "SELECT * FROM events WHERE contains(?, name)", []string{"a", "b"})
```

I get the following error with go 1.14:
`converting argument $1 type: unsupported type []string, a slice of string`

It looks like your driver needs to implement [sql/driver.NamedValueChecker](https://github.com/golang/go/blob/master/src/database/sql/driver/driver.go#L396)

nineinchnick:
Fixes #20

If #46 goes in first, I'll add some examples in the README how to pass floats and timestamps to queries, since `float64` and `time.Time` is not supported.

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,7 +190,30 @@ https://user@localhost:8443?session_properties=query_max_run_time=10m,query_prio
 
 ## Data types
 
-The driver supports most Trino data types, except:
+### Query arguments
+
+When passing arguments to queries, the driver supports the following Go data types:
+* integers
+* `bool`
+* `string`
+* slices
+
+It's not yet possible to pass:
+* `nil`
+* `float32` or `float64`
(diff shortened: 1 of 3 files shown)


losipiuk:
> If https://github.com/trinodb/trino-go-client/pull/46 goes in first, I'll add some examples in the README how to pass floats and timestamps to queries, since float64 and time.Time is not supported.

Went first :)

nineinchnick:
I added docs and it's ready to go too.

losipiuk (approved):

PR #49, opened 2022-07-13
Use GoReleaser to create GitHub releases with a changelog

nineinchnick:
Use GoReleaser because it creates nice changelogs. I used it recently in https://github.com/nineinchnick/trino-postgresql-gateway/

I'm not sure if we want the workflows to automatically bump versions and create tags. Is this new workflow self-documenting enough so we'll remember that we only need to create a tag to cut a new release?

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,36 @@
+name: release
+
+on:
+  push:
+    # run only against tags
+    tags:
+      - '*'
+
+permissions:
+  contents: read
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Fetch all tags
+        run: git fetch --force --tags
+
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: 1.18
+          cache: true
+
+      - name: Run GoReleaser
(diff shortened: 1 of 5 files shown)


losipiuk:
>I'm not sure if we want the workflows to automatically bump versions and create tags.

Probably not.

> Is this new workflow self-documenting enough so we'll remember that we only need to create a tag to cut a new release?

It would be nice to put in the readme that the process of release is to create and push a tag. And then a release flow will run automatically.

nineinchnick:
> It would be nice to put in the readme that the process of release is to create and push a tag. And then a release flow will run automatically.

On it, but maybe in `CONTRIBUTING.md`? I think `README` should be addressed only to users of the driver, not devs/maintainers.

losipiuk:
> > It would be nice to put in the readme that the process of release is to create and push a tag. And then a release flow will run automatically.
> 
> On it, but maybe in `CONTRIBUTING.md`? I think `README` should be addressed only to users of the driver, not devs/maintainers.

Fine. But i think it should still be referenced from top-level readme. Otherwise I have little faith people would find it (I would not :P)

nineinchnick:
@losipiuk I added a reference at the end of the README. PTAL

losipiuk (approved):

losipiuk:
> @losipiuk I added a reference at the end of the README. PTAL

Thanks @nineinchnick

PR #53, opened 2022-07-26
Convert recursion in `fetch` into loop

### LINKED ISSUE #52: goroutine stack exceeds 1000000000-byte limit while fetching
### issue

Query finishes in Trino Web UI, and golang server fails due to stack overflow error.
I assume error happens here at recursive fetch function.
- https://github.com/trinodb/trino-go-client/blob/master/trino/trino.go#L981

### error msg

```
runtime: goroutine stack exceeds 1000000000-byte limit
fatal error: stack overflow

runtime stack:
runtime.throw(0x1169e18, 0xe)
        /usr/local/go/src/runtime/panic.go:774 +0x72
runtime.newstack()
        /usr/local/go/src/runtime/stack.go:1046 +0x6e9
runtime.morestack()
        /usr/local/go/src/runtime/asm_amd64.s:449 +0x8f

goroutine 2676910 [running]:
reflect.Value.Type(0x111d820, 0xc5194c8fc0, 0x199, 0x0, 0x0)
        /usr/local/go/src/reflect/value.go:1874 +0x1a4 fp=0xc2cb7b8310 sp=0xc2cb7b8308 pc=0x4b8004
encoding/json.indirect(0x111d820, 0xc5194c8fc0, 0x199, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
        /usr/local/go/src/encoding/json/decode.go:498 +0x7c4 fp=0xc2cb7b8390 sp=0xc2cb7b8310 pc=0x511f44
encoding/json.(*decodeState).object(0xc5193e7208, 0x111d820, 0xc5194c8fc0, 0x199, 0x1166e11, 0xb)
        /usr/local/go/src/encoding/json/decode.go:663 +0x62 fp=0xc2cb7b8620 sp=0xc2cb7b8390 pc=0x512d12
encoding/json.(*decodeState).value(0xc5193e7208, 0x111d820, 0xc5194c8fc0, 0x199, 0x111d820, 0xc5194c8fc0)
        /usr/local/go/src/encoding/json/decode.go:429 +0x6d fp=0xc2cb7b8688 sp=0xc2cb7b8620 pc=0x5114cd
encoding/json.(*decodeState).array(0xc5193e7208, 0xf756c0, 0xc5194c8e88, 0x197, 0xc5193e7230, 0x5b)
        /usr/local/go/src/encoding/json/decode.go:616 +0x1ac fp=0xc2cb7b8770 sp=0xc2cb7b8688 pc=0x51218c
encoding/json.(*decodeState).value(0xc5193e7208, 0xf756c0, 0xc5194c8e88, 0x197, 0xf756c0, 0xc5194c8e88)
        /usr/local/go/src/encoding/json/decode.go:419 +0xfd fp=0xc2cb7b87d8 sp=0xc2cb7b8770 pc=0x51155d
encoding/json.(*decodeState).object(0xc5193e7208, 0x111d820, 0xc5194c8e10, 0x199, 0xc5193e7230, 0x7b)
        /usr/local/go/src/encoding/json/decode.go:823 +0x1392 fp=0xc2cb7b8a68 sp=0xc2cb7b87d8 pc=0x514042
encoding/json.(*decodeState).value(0xc5193e7208, 0x111d820, 0xc5194c8e10, 0x199, 0x111d820, 0xc5194c8e10)
        /usr/local/go/src/encoding/json/decode.go:429 +0x6d fp=0xc2cb7b8ad0 sp=0xc2cb7b8a68 pc=0x5114cd
encoding/json.(*decodeState).array(0xc5193e7208, 0xf756c0, 0xc5194c8bb8, 0x197, 0xc5193e7230, 0x5b)
        /usr/local/go/src/encoding/json/decode.go:616 +0x1ac fp=0xc2cb7b8bb8 sp=0xc2cb7b8ad0 pc=0x51218c
encoding/json.(*decodeState).value(0xc5193e7208, 0xf756c0, 0xc5194c8bb8, 0x197, 0xf756c0, 0xc5194c8bb8)
        /usr/local/go/src/encoding/json/decode.go:419 +0xfd fp=0xc2cb7b8c20 sp=0xc2cb7b8bb8 pc=0x51155d
encoding/json.(*decodeState).object(0xc5193e7208, 0x111d820, 0xc5194c8b40, 0x199, 0x1166e11, 0xb)
```

### sql

- I modified sql for security/privacy issue.
- result has 17,666,708 lines of rows, approx 10mb x 20 partition size in hdfs.

```sql
with topic_nx as (
	select refine_query
	from dbdb.tableA
	where gdid in (select coll_gdid
	                from dbdb.tableC
	               where log_date = '2022-07-04'
	                 and id like '...%')
	group by 1
	), web_top1_click as (
	select *
	FROM dbdb.tableB
	WHERE log_date between '2022-06-01' and '2022-06-30'
	and REFINE_QUERY IN (SELECT REFINE_QUERY
	                        FROM TOPIC_NX )
	and pageid in (select pageid
	                 from dbdb.tableA
	                group by 1)
	)

	select a.refine_query
	   , avg(case when b.qc is null then 0 else b.qc end) as Qc
	   , avg(case when c.cc is null then 0 else c.cc end) as trgt_topic_cc
	   , avg(case when d.web_cc is null then 0 else d.web_cc end) as web_cc
	   , avg(case when d.all_web_cc is null then 0 else d.all_web_cc end) as all_web_cc
	   , avg(case when d.m_all_cc is null then 0 else d.m_all_cc end) as m_all_cc
       , '2022-07-04' as log_date
	from topic_nx a
	left join (
	     select LOG_DATE, REFINE_QUERY, SUM(qc) AS QC
	       FROM dbdb.tableA
	      WHERE log_date between '2022-06-01' and '2022-06-30'
	        and REFINE_QUERY IN (SELECT REFINE_QUERY
	                               FROM TOPIC_NX
	                              group by 1)
	       group by 1,2) b
	  on a.refine_query = b.refine_query
	left join (
	     select LOG_DATE, REFINE_QUERY, SUM(cc) AS cC
	       FROM web_top1_click
	      WHERE 1=1
	        and gdid in (select coll_gdid
	                from dbdb.tableC
	               where log_date = '2022-07-04'
	                 and id like '...%')
	       group by 1,2) c
	  on a.refine_query = c.refine_query
	left join (
	     select LOG_DATE
	          , REFINE_QUERY
	          , sum(case when substr(area, 1,3) = 'area1' then cc else 0 end) as web_cc
	          , SUM(case when substr(area, 1,3) in ('area1', 'area2', 'area3', 'area4') then cc else 0 end) AS all_web_cC 
	          , sum(cc)  as m_all_cc
	       FROM web_top1_click
	       group by 1,2) d
	  on a.refine_query = d.refine_query
	group by 1
```

### Steps to reproduce

1. prepare Trino server with hive connector connected
2. launch go lang server with sql above.

```go
package main

import (
	"archive/zip"
	"database/sql"
	"fmt"
	"github.com/trinodb/trino-go-client/trino"
	_ "github.com/trinodb/trino-go-client/trino"
	"net/url"
	"os"
	"regexp"
	"strings"
)

func main() {
	u, err := url.Parse("...")
	if err != nil {
		fmt.Println(err.Error())
		return
	}
	u.User = url.UserPassword("...", "...")

	trinoConfig := &trino.Config{
		ServerURI: u.String(),
		Catalog:   "...",
	}
	dsn, err := trinoConfig.FormatDSN()
	if err != nil {
		fmt.Println(err.Error())
		return
	}
	db, err := sql.Open("trino", dsn)
	if err != nil {
		fmt.Println(err.Error())
		return
	}
	query := `...`

         // error happens here
	rows, err := db.Query(query,
		sql.Named("X-Trino-User", string("...")),
		sql.Named("X-Trino-Set-Session", "NONE"),
	)

	if err != nil {
		fmt.Println(err.Error())
		return
	}

	path, err := os.Getwd()
	if err != nil {
		fmt.Println(err.Error())
		return
	}

	filePath := path + "_trino.zip"
	_ = os.Remove(filePath) // delete before file.
	fi, err := os.OpenFile(filePath, os.O_WRONLY|os.O_CREATE, 0644)
	if err != nil {
		fmt.Println(err.Error())
		return
	}

	defer fi.Close()
	zw := zip.NewWriter(fi)
	defer zw.Close()
	iw, err := zw.Create("result.tsv")
	if err != nil {
		fmt.Println(err.Error())
		return
	}

	var header []string
	columnNames, _ := rows.Columns()
	columnTypes, err := rows.ColumnTypes()
	if err != nil {
		fmt.Println(err.Error())
		return
	}

	var results []interface{}
	var line int64
	for i := range columnNames {
		dtName := columnTypes[i].DatabaseTypeName()
		results = append(results, createType(dtName))
		header = append(header, fmt.Sprintf("%s:%s", columnNames[i], dtName))
	}

	if _, err = iw.Write([]byte(strings.Join(header, "\t") + "\n")); err != nil {
		fmt.Println(err.Error())
		return
	}

	if rows != nil {
		// Line default value is 0, but nevertheless re-assign
		line = 0
		for rows.Next() {
			err := rows.Scan(results...)
			if err != nil {
				fmt.Println(err.Error())
			}
      }
      ...

```

mdesmet:
Fixes #52


trino/trino.go
```diff
@@ -948,56 +948,62 @@ func (qr *driverRows) fetch(allowEOF bool) error {
...
-	req, err := qr.stmt.conn.newRequest("GET", qr.nextURI, nil, hs)
-	if err != nil {
-		return err
-	}
-	resp, err := qr.stmt.conn.roundTrip(qr.ctx, req)
-	if err != nil {
-		if qr.ctx.Err() == context.Canceled {
-			qr.Close()
+	for qr.nextURI != "" {
+		req, err := qr.stmt.conn.newRequest("GET", qr.nextURI, nil, hs)
```
nineinchnick, line 952: Maybe the loop body should be extracted into another private function, it's pretty big.

Also do you think you could write a test for reading a large number of rows? I wonder if it wouldn't take too long to run.

trino/trino.go
```diff
@@ -946,6 +946,48 @@ func (qr *driverRows) fetch(allowEOF bool) error {
...
+					return fmt.Errorf("error decoding column type signature: %w", err)
+				}
+			}
+			err = qr.initColumns(&qresp)
+			if err != nil {
+				return err
+			}
+		}
+
+		if len(qr.data) > 0 {
```
nineinchnick, line 983: This is always true, do a return here.

trino/trino.go
```diff
@@ -946,6 +946,48 @@ func (qr *driverRows) fetch(allowEOF bool) error {
...
+				return qr.err
+			}
+		}
+		if qr.columns == nil && len(qresp.Columns) > 0 {
+			for i := range qresp.Columns {
+				err = unmarshalArguments(&(qresp.Columns[i].TypeSignature))
+				if err != nil {
+					return fmt.Errorf("error decoding column type signature: %w", err)
+				}
+			}
```
mdesmet, line 976: This code seems better placed in the `initColumns` method. WDYT?

trino/trino.go
mdesmet: We tried adding the test but failed to make it run in an acceptable timeframe.

As the test is very specific to the code fix, it is not really beneficial to have it included in the test suite, but at least we have validated that the code indeed fixes the stack overflow error.

```
// This test ensures that the fetch method is not generating stack overflow errors.
// === RUN   TestFetchNoStackOverflow
// runtime: goroutine stack exceeds 1000000000-byte limit
// runtime: sp=0x14037b00390 stack=[0x14037b00000, 0x14057b00000]
// fatal error: stack overflow
func TestFetchNoStackOverflow(t *testing.T) {
	if testing.Short() {
		t.Skip("Skipping test in short mode.")
	}
	count := 0
	var buf *bytes.Buffer
	var ts *httptest.Server
	ts = httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		if count <= 999999 {
			if buf == nil {
				buf = new(bytes.Buffer)
				json.NewEncoder(buf).Encode(&stmtResponse{
					NextURI: ts.URL + "/v1/statement/20210817_140827_00000_arvdv/1",
				})
			}
			w.WriteHeader(http.StatusOK)
			w.Write(buf.Bytes())
			count++
			return
		}
		w.WriteHeader(http.StatusOK)
		json.NewEncoder(w).Encode(&stmtResponse{
			Error: stmtError{
				ErrorName: "TEST",
			},
		})
	}))

	db, err := sql.Open("trino", ts.URL)
	require.NoError(t, err)

	t.Cleanup(func() {
		assert.NoError(t, db.Close())
	})

	_, err = db.Query("SELECT 1")
	assert.IsTypef(t, new(ErrQueryFailed), err, "unexpected error: %w", err)
}
```

nineinchnick (approved):

nineinchnick:
@losipiuk PTAL

trino/trino.go
nineinchnick: If you add `debug.SetMaxStack(20 * 1024)` at the beginning of the test and lower the number of responses from 999999 to only 99, the test fails on master and passes in this branch and takes about 200ms, which IMO is acceptable.
Using `debug.SetMemoryLimit(100 * 1024 * 1024)` from Go 1.19 does not have any effect as the issue is about the stack size.

nineinchnick: Also, this is going to be handy to use as a benchmark. I did some memory profiling, and we're doing too many allocations, making the Go driver fall behind Python and Java. There's also an issue with loading the time zones when parsing timestamps. This is out of scope of this PR, but this test is valuable.

mdesmet: > If you add debug.SetMaxStack(20 * 1024) at the beginning of the test and lower the number of responses from 999999 to only 99, the test fails on master and passes in this branch and takes about 200ms, which IMO is acceptable.

For some reason it locally failed on me in both branches with these numbers.  `debug.SetMaxStack(50 * 1024)` and 50 requests did fail on master but work on this branch. Just to be sure maybe retest this config on your side (to make sure we are not introducing a flaky test).

trino/trino_test.go
```diff
@@ -686,6 +688,48 @@ func TestQueryFailure(t *testing.T) {
 	assert.IsTypef(t, new(ErrQueryFailed), err, "unexpected error: %w", err)
 }
 
+// This test ensures that the fetch method is not generating stack overflow errors.
+// === RUN   TestFetchNoStackOverflow
+// runtime: goroutine stack exceeds 1000000000-byte limit
+// runtime: sp=0x14037b00390 stack=[0x14037b00000, 0x14057b00000]
+// fatal error: stack overflow
+func TestFetchNoStackOverflow(t *testing.T) {
+	debug.SetMaxStack(50 * 1024)
```
losipiuk, line 697: don't you need to reset it at the end of the method.
Also what if multiple tests are running at the same time.

Sorry for noob questions :)

nineinchnick: Those are valid questions. I ran all the tests with this one as the first and everything works. I assume all other tests don't require a bigger stack. We can either:
* revert this at the end of the test
* set it in `TestMain`

Tests don't run in parallel by default and it requires an explicit flag `-parallel`. I tried it now and some integration tests fail when checking for query status. I assume they're not written to support parallel execution, that is they don't use unique identifiers.

losipiuk: If tests are not built to run in parallel then I would revert at the end of test via `defer`.

mdesmet: That's done already :)

```
previousSetting := debug.SetMaxStack(50 * 1024)
defer debug.SetMaxStack(previousSetting)
```

Note that i also had to increase the timeout of the tests on the CI.

nineinchnick:
@mdesmet can you rebase?

losipiuk (approved):

losipiuk:
@nineinchnick are you good with this PR?

nineinchnick:
Yes, lgtm

PR #57, opened 2022-09-21
Concurrent requests

nineinchnick:
Introduce two goroutines for making HTTP requests and decoding the JSON payload concurrently. This brings only a minor improvement (see benchmarks below) unless the server can send the `nextUri` as an HTTP header, so the next request can be made immediately, without waiting to decode the payload. 

Before:
```
% go test -v -race -timeout 1m ./... -run=X -bench=. -no_cleanup
goos: darwin
goarch: arm64
pkg: github.com/trinodb/trino-go-client/trino
BenchmarkQuery
BenchmarkQuery-10    	      1	55366793167 ns/op
PASS
ok  	github.com/trinodb/trino-go-client/trino	67.287s
go test -v -race -timeout 1m ./... -run=X -bench=. -no_cleanup  51.59s user 0.88s system 77% cpu 1:07.69 total
```

After:
```
% go test -v -race -timeout 1m ./... -run=X -bench=. -no_cleanup
goos: darwin
goarch: arm64
pkg: github.com/trinodb/trino-go-client/trino
BenchmarkQuery
BenchmarkQuery-10    	      1	48234415166 ns/op
PASS
ok  	github.com/trinodb/trino-go-client/trino	48.676s
go test -v -race -timeout 1m ./... -run=X -bench=. -no_cleanup  51.94s user 0.89s system 107% cpu 49.096 total
```

### DIFF
--- a/trino/trino.go
+++ b/trino/trino.go
@@ -547,11 +547,15 @@ func newErrQueryFailedFromResponse(resp *http.Response) *ErrQueryFailed {
 }
 
 type driverStmt struct {
-	conn    *Conn
-	query   string
-	user    string
-	statsCh chan QueryProgressInfo
-	doneCh  chan struct{}
+	conn           *Conn
+	query          string
+	user           string
+	nextURIs       chan string
+	httpResponses  chan *http.Response
+	queryResponses chan queryResponse
+	statsCh        chan QueryProgressInfo
+	errors         chan error
+	doneCh         chan struct{}
 }
 
 var (
@@ -563,12 +567,26 @@ var (
 
 // Close closes statement just before releasing connection
 func (st *driverStmt) Close() error {
-	if st.doneCh != nil {
-		close(st.doneCh)
+	if st.doneCh == nil {
+		return nil
 	}
+	close(st.doneCh)
 	if st.statsCh != nil {
 		<-st.statsCh
+		st.statsCh = nil
 	}
+	go func() {
+		// drain errors chan to allow goroutines to write to it
+		for range st.errors {
+		}
+	}()
+	for range st.queryResponses {
+	}
+	for range st.httpResponses {
+	}
+	close(st.nextURIs)
+	close(st.errors)
+	st.doneCh = nil
 	return nil
 }
 
@@ -596,7 +614,7 @@ func (st *driverStmt) ExecContext(ctx context.Context, args []driver.NamedValue)
 	}
 	// consume all results, if there are any
 	for err == nil {
-		err = rows.fetch(true)
+		err = rows.fetch()
 	}
 
 	if err != nil && err != io.EOF {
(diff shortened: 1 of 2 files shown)


nineinchnick:
Here's the remaining work required to handle the `nextUri` send in the header: https://github.com/nineinchnick/trino-go-client/compare/concurrent-requests...nineinchnick:trino-go-client:concurrent-requests-header

trino/trino_test.go
```diff
@@ -1385,3 +1385,29 @@ func TestSlice3TypeConversion(t *testing.T) {
 		})
 	}
 }
+
+func BenchmarkQuery(b *testing.B) {
+	c := &Config{
+		ServerURI:         *integrationServerFlag,
+		SessionProperties: map[string]string{"query_priority": "1"},
```
losipiuk, line 1392: what is it for?

trino/trino.go
```diff
@@ -43,12 +43,11 @@
 //
 // The driver should be used via the database/sql package:
 //
-//  import "database/sql"
-//  import _ "github.com/trinodb/trino-go-client/trino"
-//
-//  dsn := "http://user@localhost:8080?catalog=default&schema=test"
-//  db, err := sql.Open("trino", dsn)
+//	import "database/sql"
```
losipiuk, line 46: nit: unrelated. Extract to separate commit

trino/trino_test.go
nineinchnick: I'm not sure, I copied this config from other tests :-)

nineinchnick:
I need to add a test for a query with a larger number of result rows to make sure we're not dropping any rows.

trino/trino.go
```diff
@@ -136,6 +135,8 @@ const (
 	kerberosRealmConfig      = "KerberosRealm"
 	kerberosConfigPathConfig = "KerberosConfigPath"
 	SSLCertPathConfig        = "SSLCertPath"
+
+	CHAN_SIZE = 10
```
nineinchnick, line 139: We might want to try a different value here, maybe 2 is sufficient.

losipiuk: How big a single response can be?

nineinchnick: It's 1MB by default: https://github.com/trinodb/trino/blob/master/core/trino-main/src/main/java/io/trino/server/protocol/ExecutingStatementResource.java#L80

losipiuk: Then I guess 10 is fine. Unless it does not improve anything compared to 2. Then we can go lower. I guess it will depend on number of cores we have and if we are actually able to read data quickly enough to saturate CPU - i kinda doubt that :)

nineinchnick: The benchmark returns the same results when using unbuffered channels, so I removed this. The benchmark doesn't do any data processing, so no actual application can be faster than that and won't benefit from additional buffering. Storing both a single HTTP response and the decoded payload is enough buffering.

losipiuk (approved):

nineinchnick:
@losipiuk it's ready to go, thanks!

PR #61, opened 2022-11-14
Change the fieldName field type from string to rowFieldName struct

Sunwoo-Shin:
### Situation

If a name is assigned to a ROW type, the following error occurs.

스크린샷 2022-11-14 오후 11 17 38

The fieldName field is not in the form of a simple string, but in the form of JSON nested objects with a field called name.

```json
{
    "rawType":"row",
    "arguments":[
        {
            "kind":"NAMED_TYPE",
            "value":{
                "fieldName":{"name":"x"},
                "typeSignature":{"rawType":"varchar","arguments":[{"kind":"LONG","value":2147483647}]}
            }
        },
        {
            "kind":"NAMED_TYPE",
            "value":{
                "fieldName":{"name":"y"},
                "typeSignature":{"rawType":"double","arguments":[]}
            }
        }
    ]
}
```

### Reference Source
* NamedTypeSignature : https://github.com/trinodb/trino/blob/fe608f2723842037ff620d612a706900e79c52c8/core/trino-spi/src/main/java/io/trino/spi/type/NamedTypeSignature.java#L27
* RowFieldName: https://github.com/trinodb/trino/blob/fe608f2723842037ff620d612a706900e79c52c8/core/trino-spi/src/main/java/io/trino/spi/type/RowFieldName.java#L25

### DIFF
--- a/trino/trino.go
+++ b/trino/trino.go
@@ -1070,8 +1070,12 @@ type queryColumn struct {
 type queryData []interface{}
 
 type namedTypeSignature struct {
-	FieldName     string        `json:"fieldName"`
+	FieldName     rowFieldName  `json:"fieldName"`
 	TypeSignature typeSignature `json:"typeSignature"`
+}
+
+type rowFieldName struct {
+	Name string `json:"name"`
 }
 
 type typeSignature struct {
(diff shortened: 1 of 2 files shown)


nineinchnick:
@losipiuk PTAL

losipiuk (approved):

PR #62, opened 2022-12-04
Support offset timezone, times do not have spaces

JPMoresmau:
As seen in https://trino.io/docs/current/language/types.html#time-with-time-zone, numeric offsets are supported and there is not always a space before the time zone. This PR fixes that.
I have sent a CLA a few days ago, hope this is oK.

### DIFF
--- a/trino/trino.go
+++ b/trino/trino.go
@@ -1869,6 +1869,12 @@ var timeLayouts = []string{
 	"2006-01-02 15:04:05.000",
 }
 
+// Layout for time and timestamp WITH time zone.
+var timeLayoutsTZ = []string{
+	"15:04:05.000 -07:00",
+	"2006-01-02 15:04:05.000 -07:00",
+}
+
 func scanNullTime(v interface{}) (NullTime, error) {
 	if v == nil {
 		return NullTime{}, nil
@@ -1880,6 +1886,19 @@ func scanNullTime(v interface{}) (NullTime, error) {
 	vparts := strings.Split(vv, " ")
 	if len(vparts) > 1 && !unicode.IsDigit(rune(vparts[len(vparts)-1][0])) {
 		return parseNullTimeWithLocation(vv)
+	}
+	// Time literals may not have spaces before the timezone.
+	if strings.ContainsRune(vv, '+') {
+		return parseNullTimeWithLocation(strings.Replace(vv, "+", " +", 1))
+	}
+	hyphenCount := strings.Count(vv, "-")
+	// We need to ensure we don't treat the hyphens in dates as the minus offset sign.
+	// So if there's only one hyphen or more than 2, we have a negative offset.
+	if hyphenCount == 1 || hyphenCount > 2 {
+		// We add a space before the last hyphen to parse properly.
+		i := strings.LastIndex(vv, "-")
+		timestamp := vv[:i] + strings.Replace(vv[i:], "-", " -", 1)
+		return parseNullTimeWithLocation(timestamp)
 	}
 	return parseNullTime(vv)
 }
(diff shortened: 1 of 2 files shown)


trino/trino.go
```diff
@@ -1881,6 +1887,17 @@ func scanNullTime(v interface{}) (NullTime, error) {
 	if len(vparts) > 1 && !unicode.IsDigit(rune(vparts[len(vparts)-1][0])) {
 		return parseNullTimeWithLocation(vv)
 	}
+	// Time literals may not have spaces before the timezone.
+	vparts = strings.Split(vv, "+")
```
nineinchnick, line 1891: Would checking if `vv` contains a `+` be enough (using `strings.Contains()`)? It would help avoid allocating the `vparts` slice that's not used otherwise.

trino/trino.go
```diff
@@ -1903,7 +1920,15 @@ func parseNullTimeWithLocation(v string) (NullTime, error) {
 	}
 	stamp, location := v[:idx], v[idx+1:]
 	loc, err := time.LoadLocation(location)
+	// Not a named location.
 	if err != nil {
```
nineinchnick, line 1924: Is it possible to inspect the `err` to see if it failed to parse the value or is it an actual problem with loading TZ data? Or since we already extracted the location from the string, maybe we could check its format first?

trino/trino.go
JPMoresmau: Good idea, done.

trino/trino.go
JPMoresmau: Instead of looking at `err` we can just look if location looks like an offset.

trino/trino.go
JPMoresmau: Well, it's more complicated because we have hyphens in the date part, so a simple contains is not sufficient. I now use Count we should reduce the allocations. Thanks!

trino/trino.go
```diff
@@ -1881,6 +1887,16 @@ func scanNullTime(v interface{}) (NullTime, error) {
...
 		return parseNullTimeWithLocation(vv)
 	}
+	// Time literals may not have spaces before the timezone.
+	if strings.ContainsRune(vv, '+') {
+		return parseNullTimeWithLocation(strings.Replace(vv, "+", " +", 1))
+	}
+	hiphenCount := strings.Count(vv, "-")
+	// We need to ensure we don't treat the hyphens in dates as the minus offset sign.
+	// So if there's only one hyphen or more than 2, we have a negative offset.
+	if hiphenCount == 1 || hiphenCount > 2 {
```
nineinchnick, line 1897: nit: typo, also I'm not a native speaker and had to check if it's better to call this minus or hyphen, and TIL about [hyphen-minus](https://en.wikipedia.org/wiki/Hyphen-minus) 😄  
```suggestion
	hyphenCount := strings.Count(vv, "-")
	// We need to ensure we don't treat the hyphens in dates as the minus offset sign.
	// So if there's only one hyphen or more than 2, we have a negative offset.
	if hyphenCount == 1 || hyphenCount > 2 {
```

JPMoresmau:
@cla-bot check

JPMoresmau:
Sorry, is anything more needed to get this PR approved?

nineinchnick (approved):
LGTM, thanks!

JPMoresmau:
Would it be a problem to merge this and do a release with it, like was done for my other PR? I kinda like need this functionality for the Trino support we're building... Thanks!

nineinchnick:
I think so, @losipiuk can you handle this?

losipiuk:
@JPMoresmau please squash the commits.

JPMoresmau:
@losipiuk done!

trino/trino_test.go
```diff
@@ -999,6 +999,18 @@ func TestTypeConversion(t *testing.T) {
...
+			RawType:                    "time with time zone",
+			ResponseUnmarshalledSample: "01:02:03.000 +03:00",
+			ExpectedGoValue:            time.Date(0, 1, 1, 1, 2, 3, 0, time.FixedZone("", 3*3600)),
+		},
+		{
+			DataType:                   "time with time zone",
+			RawType:                    "time with time zone",
+			ResponseUnmarshalledSample: "01:02:03.000+03:00",
+			ExpectedGoValue:            time.Date(0, 1, 1, 1, 2, 3, 0, time.FixedZone("", 3*3600)),
+		},
```
losipiuk, line 1013: can you please also add tests with `-` (with and without space)

trino/trino_test.go
```diff
@@ -1011,6 +1023,12 @@ func TestTypeConversion(t *testing.T) {
 			ResponseUnmarshalledSample: "2017-07-10 01:02:03.000 UTC",
 			ExpectedGoValue:            time.Date(2017, 7, 10, 1, 2, 3, 0, utc),
 		},
+		{
+			DataType:                   "timestamp with time zone",
+			RawType:                    "timestamp with time zone",
+			ResponseUnmarshalledSample: "2017-07-10 01:02:03.000 +03:00",
+			ExpectedGoValue:            time.Date(2017, 7, 10, 1, 2, 3, 0, time.FixedZone("", 3*3600)),
+		},
```
losipiuk, line 1031: same here. (also -, with and without space)

trino/trino.go
```diff
@@ -1869,6 +1869,12 @@ var timeLayouts = []string{
 	"2006-01-02 15:04:05.000",
 }
 
+// Layout for time and timestamp WITH time zone.
+var timeLayoutsTZ = []string{
```
losipiuk, line 1873: change commit message to
```
Add support for offset timezones for time and timestamp with timezone
```

losipiuk:
new release should be on its way. Thanks

PR #64, opened 2022-12-27
Support times better

JPMoresmau:
Add support for full precision of subseconds, so `PARAMETRIC_DATETIME` works properly.

Add type wrappers in the manner of `Numeric` for the various date/time data types, so we can avoid casts and not leak the accepted formats to the client.

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,12 +198,17 @@ When passing arguments to queries, the driver supports the following Go data typ
 * `string`
 * slices
 * `trino.Numeric` - a string representation of a number
+* `time.Time` - passed to Trino as a timestamp with a time zone
+* the result of `trino.Date(year, month, day)` - passed to Trino as a date
+* the result of `trino.Time(hour, minute, second, nanosecond)` - passed to Trino as a time without a time zone
+* the result of `trino.TimeTz(hour, minute, second, nanosecond, location)` - passed to Trino as a time with a time zone
+* the result of `trino.Timestamp(year, month, day, hour, minute, second, nanosecond)` - passed to Trino as a timestamp without a time zone
 
 It's not yet possible to pass:
 * `nil`
 * `float32` or `float64`
 * `byte`
-* `time.Time` or `time.Duration`
+* `time.Duration`
 * `json.RawMessage`
 * maps
 
@@ -215,7 +220,11 @@ SELECT * FROM table WHERE col_double = cast(? AS DOUBLE) OR col_timestamp = CAST
 ### Response rows
 
 When reading response rows, the driver supports most Trino data types, except:
-* time and timestamps with precision - all time types are returned as `time.Time`
+* time and timestamps with precision - all time types are returned as `time.Time`.
+  All precisions up to nanoseconds (`TIMESTAMP(9)` or `TIME(9)`) are supported (since
+  this is the maximum precision Golang's `time.Time` supports). If a query returns columns
+  defined with a greater precision, use `CAST` to reduce the returned precision, or convert the
+  value to a string that then can be parsed manually.
 * `DECIMAL` - returned as string
 * `IPADDRESS` - returned as string
 * `INTERVAL YEAR TO MONTH` and `INTERVAL DAY TO SECOND` - returned as string
--- a/trino/serial.go
+++ b/trino/serial.go
@@ -35,6 +35,65 @@ func (e UnsupportedArgError) Error() string {
 // If another string format is used it will error to serialise
 type Numeric string
 
+// trinoDate represents a Date type in Trino.
+type trinoDate struct {
+	year  int
+	month time.Month
+	day   int
+}
+
+// Date creates a representation of a Trino Date type.
+func Date(year int, month time.Month, day int) trinoDate {
+	return trinoDate{year, month, day}
+}
+
+// trinoTime represents a Time type in Trino.
+type trinoTime struct {
+	hour       int
+	minute     int
+	second     int
+	nanosecond int
+}
+
+// Time creates a representation of a Trino Time type. To represent time with precision higher than nanoseconds, pass the value as a string and use a cast in the query.
+func Time(hour int,
+	minute int,
+	second int,
+	nanosecond int) trinoTime {
+	return trinoTime{hour, minute, second, nanosecond}
+}
+
+// trinoTimeTz represents a Time(9) With Timezone type in Trino.
+type trinoTimeTz time.Time
+
+// TimeTz creates a representation of a Trino Time(9) With Timezone type.
+func TimeTz(hour int,
+	minute int,
+	second int,
+	nanosecond int,
+	location *time.Location) trinoTimeTz {
+	// When reading a time, a nil location indicates UTC.
+	// However, passing nil to time.Date() panics.
+	if location == nil {
+		location = time.UTC
+	}
+	return trinoTimeTz(time.Date(0, 0, 0, hour, minute, second, nanosecond, location))
+}
+
+// Timestamp indicates we want a TimeStamp type WITHOUT a time zone in Trino from a Golang time.
+type trinoTimestamp time.Time
+
+// Timestamp creates a representation of a Trino Timestamp(9) type.
+func Timestamp(year int,
+	month time.Month,
+	day int,
+	hour int,
+	minute int,
+	second int,
+	nanosecond int) trinoTimestamp {
+	return trinoTimestamp(time.Date(year, month, day, hour, minute, second, nanosecond, time.UTC))
+}
+
 // Serial converts any supported value to its equivalent string for as a Trino parameter
 // See https://trino.io/docs/current/language/types.html
 func Serial(v interface{}) (string, error) {
@@ -92,9 +151,17 @@ func Serial(v interface{}) (string, error) {
 	case []byte:
 		return "", UnsupportedArgError{"[]byte"}
 
-		// time.Time and time.Duration not supported as time and date take several different formats in Trino
+	case trinoDate:
+		return fmt.Sprintf("DATE '%04d-%02d-%02d'", x.year, x.month, x.day), nil
+	case trinoTime:
+		return fmt.Sprintf("TIME '%02d:%02d:%02d.%09d'", x.hour, x.minute, x.second, x.nanosecond), nil
+	case trinoTimeTz:
+		return "TIME " + time.Time(x).Format("'15:04:05.999999999 Z07:00'"), nil
+	case trinoTimestamp:
+		return "TIMESTAMP " + time.Time(x).Format("'2006-01-02 15:04:05.999999999'"), nil
 	case time.Time:
-		return "", UnsupportedArgError{"time.Time"}
+		return "TIMESTAMP " + time.Time(x).Format("'2006-01-02 15:04:05.999999999 Z07:00'"), nil
+
 	case time.Duration:
 		return "", UnsupportedArgError{"time.Duration"}
 
--- a/trino/serial_test.go
+++ b/trino/serial_test.go
@@ -14,9 +14,16 @@
 
 package trino
 
-import "testing"
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
 
 func TestSerial(t *testing.T) {
+	paris, err := time.LoadLocation("Europe/Paris")
+	require.NoError(t, err)
 	scenarios := []struct {
 		name           string
 		value          interface{}
@@ -112,6 +119,46 @@ func TestSerial(t *testing.T) {
 			name:           "bool false",
 			value:          false,
 			expectedSerial: "false",
+		},
+		{
+			name:           "date",
+			value:          Date(2017, 7, 10),
+			expectedSerial: "DATE '2017-07-10'",
+		},
+		{
+			name:           "time without timezone",
+			value:          Time(11, 34, 25, 123456),
+			expectedSerial: "TIME '11:34:25.000123456'",
+		},
+		{
+			name:           "time with timezone",
+			value:          TimeTz(11, 34, 25, 123456, time.FixedZone("test zone", +2*3600)),
+			expectedSerial: "TIME '11:34:25.000123456 +02:00'",
+		},
+		{
+			name:           "time with timezone",
+			value:          TimeTz(11, 34, 25, 123456, nil),
+			expectedSerial: "TIME '11:34:25.000123456 Z'",
+		},
+		{
+			name:           "timestamp without timezone",
+			value:          Timestamp(2017, 7, 10, 11, 34, 25, 123456),
+			expectedSerial: "TIMESTAMP '2017-07-10 11:34:25.000123456'",
+		},
+		{
+			name:           "timestamp with time zone in Fixed Zone",
+			value:          time.Date(2017, 7, 10, 11, 34, 25, 123456, time.FixedZone("test zone", +2*3600)),
+			expectedSerial: "TIMESTAMP '2017-07-10 11:34:25.000123456 +02:00'",
+		},
+		{
+			name:           "timestamp with time zone in Named Zone",
+			value:          time.Date(2017, 7, 10, 11, 34, 25, 123456, paris),
+			expectedSerial: "TIMESTAMP '2017-07-10 11:34:25.000123456 +02:00'",
+		},
+		{
+			name:           "timestamp with time zone in UTC",
+			value:          time.Date(2017, 7, 10, 11, 34, 25, 123456, time.UTC),
+			expectedSerial: "TIMESTAMP '2017-07-10 11:34:25.000123456 Z'",
 		},
 		{
 			name:          "nil",
--- a/trino/trino.go
+++ b/trino/trino.go
@@ -624,20 +624,24 @@ func (st *driverStmt) ExecContext(ctx context.Context, args []driver.NamedValue)
 }
 
 func (st *driverStmt) CheckNamedValue(arg *driver.NamedValue) error {
-	_, ok := arg.Value.(Numeric)
-	if ok {
+	switch arg.Value.(type) {
+	case Numeric, trinoDate, trinoTime, trinoTimeTz, trinoTimestamp:
 		return nil
-	}
-	if reflect.TypeOf(arg.Value).Kind() == reflect.Slice {
-		return nil
+	default:
+		{
+			if reflect.TypeOf(arg.Value).Kind() == reflect.Slice {
+				return nil
+			}
+
+			if arg.Name == trinoProgressCallbackParam {
+				return nil
+			}
+			if arg.Name == trinoProgressCallbackPeriodParam {
+				return nil
+			}
+		}
 	}
 
-	if arg.Name == trinoProgressCallbackParam {
-		return nil
-	}
-	if arg.Name == trinoProgressCallbackPeriodParam {
-		return nil
-	}
 	return driver.ErrSkip
 }
 
@@ -733,10 +737,11 @@ func (st *driverStmt) QueryContext(ctx context.Context, args []driver.NamedValue
 
 func (st *driverStmt) exec(ctx context.Context, args []driver.NamedValue) (*stmtResponse, error) {
 	query := st.query
-	var hs http.Header
+	hs := make(http.Header)
+	// Ensure the server returns timestamps preserving their precision, without truncating them to timestamp(3).
+	hs.Add("X-Trino-Client-Capabilities", "PARAMETRIC_DATETIME")
 
 	if len(args) > 0 {
-		hs = make(http.Header)
 		var ss []string
 		for _, arg := range args {
 			if arg.Name == trinoProgressCallbackParam {
@@ -1279,7 +1284,15 @@ func newTypeConverter(typeName string, signature typeSignature) (*typeConverter,
 			}
 			result.scale = newOptionalInt64(signature.Arguments[1].long)
 		}
+	case "time", "time with time zone", "timestamp", "timestamp with time zone":
+		if len(signature.Arguments) > 0 {
+			if signature.Arguments[0].Kind != KIND_LONG {
+				return nil, ErrInvalidResponseType
+			}
+			result.precision = newOptionalInt64(signature.Arguments[0].long)
+		}
 	}
+
 	return result, nil
 }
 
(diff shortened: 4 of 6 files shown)


trino/trino_test.go
```diff
@@ -765,6 +765,106 @@ func TestQueryColumns(t *testing.T) {
...
+	t.Cleanup(func() {
+		assert.NoError(t, db.Close())
+	})
+
+	rows, err := db.Query(`SELECT
+  cast(current_time as time(6)) AS timep,
+  cast(current_time as time(6) with time zone) AS timeptz,
+  cast(current_time as timestamp(6)) AS tsp,
+  cast(current_time as timestamp(6) with time zone) AS tsptz`,
+		sql.Named("X-Trino-Client-Capabilities", "PARAMETRIC_DATETIME"))
```
nineinchnick, line 789: Can this be enabled by default? This would simplify usage and reduce cognitive load - I don't remember how exactly this parameter behaves and would have to check Trino sources or other clients. I wouldn't expect users of this driver to know this.

Whatever we decide here, the README also should be updated.

trino/trino_test.go
```diff
@@ -765,6 +765,106 @@ func TestQueryColumns(t *testing.T) {
...
+	require.NoError(t, err, "Failed executing query")
+	assert.NotNil(t, rows)
+
+	columns, err := rows.Columns()
+	require.NoError(t, err, "Failed reading result columns")
+
+	assert.Equal(t, 4, len(columns), "Expected 4 result column")
```
nineinchnick, line 796: Is this necessary? Wouldn't `assert.Equal` called on a slice verify this?

trino/trino_test.go
```diff
@@ -765,6 +765,106 @@ func TestQueryColumns(t *testing.T) {
...
+			0,
+			0,
+			false,
+			0,
+			reflect.TypeOf(sql.NullTime{}),
+		},
+		{
+			"TIMESTAMP WITH TIME ZONE",
+			false,
+			0,
```
nineinchnick, line 850: Shouldn't this be non-zero?

trino/trino_test.go
```diff
@@ -765,6 +765,106 @@ func TestQueryColumns(t *testing.T) {
...
+
+	t.Cleanup(func() {
+		assert.NoError(t, db.Close())
+	})
+
+	rows, err := db.Query(`SELECT
+  cast(current_time as time(6)) AS timep,
+  cast(current_time as time(6) with time zone) AS timeptz,
+  cast(current_time as timestamp(6)) AS tsp,
+  cast(current_time as timestamp(6) with time zone) AS tsptz`,
```
nineinchnick, line 788: What's the expected behavior for a high precision, like 12? If we can't store such values and they would be truncated, an error should be raised, or they should be returned as strings.

trino/trino_test.go
```diff
@@ -1070,6 +1206,42 @@ func TestTypeConversion(t *testing.T) {
...
+		{
+			DataType:                   "timestamp with time zone",
+			RawType:                    "timestamp with time zone",
+			ResponseUnmarshalledSample: "2017-07-10 01:02:03.123456789 UTC",
+			ExpectedGoValue:            time.Date(2017, 7, 10, 1, 2, 3, 123456789, utc),
+		},
+		{
+			DataType:                   "timestamp with time zone",
+			RawType:                    "timestamp with time zone",
+			ResponseUnmarshalledSample: "2017-07-10 01:02:03.123456789 +03:00",
```
nineinchnick, line 1224: Can the Trino server send a geographical TZ, like `Europe/Warsaw`, not just an offset?

trino/trino.go
```diff
@@ -1863,16 +1867,21 @@ func (s *NullSlice3Float64) Scan(value interface{}) error {
...
 var timeLayouts = []string{
 	"2006-01-02",
-	"15:04:05.000",
-	"2006-01-02 15:04:05.000",
+	"15:04:05.999999999999",
+	"2006-01-02 15:04:05.999999999999",
 }
 
 // Layout for time and timestamp WITH time zone.
+// Trino can support up to 12 digits sub second precision.
```
nineinchnick, line 1880: Can we really store picoseconds in a `time.Time`? [The docs](https://pkg.go.dev/time#Time) say:
> A Time represents an instant in time with nanosecond precision.

nineinchnick:
Looks like this would resolve #13

trino/serial.go
```diff
@@ -35,6 +35,21 @@ func (e UnsupportedArgError) Error() string {
 // If another string format is used it will error to serialise
 type Numeric string
 
+// Date indicates we want a Date type in Trino from a Golang time.
+type Date time.Time
```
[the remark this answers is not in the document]
JPMoresmau: OK, so instead of just a type wrapper we make Date and Time more explicit structs with only the fields they need? And keep the time wrappers for timestamps?

trino/trino_test.go
JPMoresmau: I just copied the test code from another test :-)

trino/trino.go
JPMoresmau: So but we need to be able to parse the string containing 12 digits into Go. And yes I guess the value will get truncated because Go doesn't support better than nanos

trino/trino_test.go
JPMoresmau: Yes, I think so, I'll check this works.

trino/trino_test.go
hashhar: In every client which supports this we want it to be the default behaviour - the Python client will start supporting this in next release and also defaults to always use the PARAMETRIC_DATETIME behaviour. It's a "breaking change" in some sense but IMO the benefits of always using parametric_datetime outweigh possible breakage.

JPMoresmau: I have no problem enabling this since I set it in my own code, let me know if it's OK and if there should be a mechanism to not pass it. I stumbled upon this while investigating why my timestamps were truncated to milliseconds: https://github.com/trinodb/trino/issues/5068

trino/serial.go
```diff
@@ -92,7 +107,19 @@ func Serial(v interface{}) (string, error) {
 	case []byte:
 		return "", UnsupportedArgError{"[]byte"}
 
+	case Date:
+		return "DATE " + time.Time(x).Format("'2006-01-02'"), nil
```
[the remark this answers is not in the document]
nineinchnick: You might create a `Date` somewhere from a `time.Time` that you created earlier or received from another part of a program and not know it has a non-zero time value. Consider a function like this:
```
func getOrders(since time.Time) []Order {
    rows, err := conn.Query(`SELECT * FROM orders WHERE created_at > ?`, trino.Date(since))
    // ...
}
```
This is error-prone, and the driver should not allow this. If we'd require the callers to explicitly create the Date from parts of `time.Time`, like `trino.Date(since.Year(), since.Month(), since.Day())` it's obvious to the caller that the time part is being ignored.

trino/trino_test.go
hashhar: In Python client we bit the bullet and enabled it by default with no option to fallback (since I can't think of any case where the alternative behaviour of precision always being 3 is useful).

I'd like to hear @losipiuk's opinion as well on this.

trino/serial.go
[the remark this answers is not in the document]
nineinchnick: How they store time internally would be an implementation detail. I guess it only matters if we want to support nano or picoseconds. I was thinking about how to avoid surprises when using them.

Allowing to use `time.Time` directly would be nice too.

trino/trino.go
nineinchnick: The driver should not have to parse it just to throw away picoseconds. If it cannot handle picoseconds, it's reasonable to expect queries to be written in a way to avoid returning them (either round or truncate explicitly) and return an actionable error message explaining this.

trino/serial.go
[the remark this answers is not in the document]
JPMoresmau: I have replaced the type wrappers with functions that explicitly request the various fields they contain, or make it explicit what they strip, let me know. I've also removed TimestampTz but added direct support for time.Time (since according to the doc time.Time is ALWAYS carrying a timezone, since nil location means UTC).

trino/trino_test.go
JPMoresmau: Added a couple of tests, this works.

trino/trino_test.go
JPMoresmau: We build another slice later on using that length, so I guess it's safer to verify it...

trino/trino_test.go
JPMoresmau: Added support to capture timestamp and time column precision.

trino/serial.go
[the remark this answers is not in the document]
JPMoresmau: Done

trino/serial_test.go
```diff
@@ -113,6 +116,31 @@ func TestSerial(t *testing.T) {
 			value:          false,
 			expectedSerial: "false",
 		},
+		{
+			name:           "date",
+			value:          Date(time.Date(2017, 7, 10, 0, 0, 0, 0, time.Local)),
```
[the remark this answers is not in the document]
JPMoresmau: Since the way to construct these has been changed, there is no hidden truncation

trino/trino_test.go
JPMoresmau: Enabled by default, let me know.

trino/trino_test.go
JPMoresmau: We now error if the precision is greater than nanoseconds

trino/trino.go
JPMoresmau: We now error on precision greater than nanoseconds. So reading a timestamp as `TIMESTAMP(9)` works but greater precision fails.

trino/trino_test.go
JPMoresmau: README updated to document the date/time support

trino/serial.go
```diff
@@ -35,6 +35,55 @@ func (e UnsupportedArgError) Error() string {
...
 type Numeric string
 
+// Date indicates we want a Date type in Trino from a Golang time.
+type trinoDate struct {
+	year  int
+	month time.Month
+	day   int
+}
+
+func Date(year int, month time.Month, day int) trinoDate {
```
nineinchnick, line 45: All exported functions should have comments, maybe something like:
```suggestion
// Date creates a representation of a Trino Date type.
func Date(year int, month time.Month, day int) trinoDate {
```

trino/serial.go
```diff
@@ -35,6 +35,55 @@ func (e UnsupportedArgError) Error() string {
...
+
+// Time indicates we want a Time type in Trino from a Golang time.
+type trinoTime struct {
+	hour       int
+	minute     int
+	second     int
+	nanosecond int
+}
+
+func Time(hour int,
```
nineinchnick, line 57: Same here. Also, I wonder if it would be too much to describe what to do when a higher precision is needed.
```suggestion
// Time creates a representation of a Trino Time type. To represent time with precision higher than nanoseconds, pass the value as a string and use a cast in the query.
func Time(hour int,
```

trino/serial.go
```diff
@@ -92,9 +141,18 @@ func Serial(v interface{}) (string, error) {
...
+		return fmt.Sprintf("TIME '%02d:%02d:%02d.%09d'", x.hour, x.minute, x.second, x.nanosecond), nil
+	case trinoTimeTz:
+		return "TIME " + time.Time(x).Format("'15:04:05.999999999 Z07:00'"), nil
+	case trinoTimestamp:
+		return "TIMESTAMP " + time.Time(x).Format("'2006-01-02 15:04:05.999999999'"), nil
 	case time.Time:
-		return "", UnsupportedArgError{"time.Time"}
+		return "TIMESTAMP " + time.Time(x).Format("'2006-01-02 15:04:05.999999999 Z07:00'"), nil
+
+		// TODO - support durations
```
nineinchnick, line 155: The TODO seems unnecessary, there's an informative error being returned.

trino/trino.go
```diff
@@ -733,10 +737,11 @@ func (st *driverStmt) QueryContext(ctx context.Context, args []driver.NamedValue
 
 func (st *driverStmt) exec(ctx context.Context, args []driver.NamedValue) (*stmtResponse, error) {
 	query := st.query
-	var hs http.Header
+	hs := make(http.Header)
+	// Ensure we have full date precision support.
```
nineinchnick, line 741: I checked [the docs](https://trino.io/docs/current/develop/client-protocol.html) to see if we could copy the description but it doesn't mention this header so we can't reference it here. How about this:
```suggestion
	// Ensure the server returns timestamps preserving their precision, without truncating them to timestamp(3).
```

trino/serial_test.go
```diff
@@ -14,9 +14,16 @@
...
-import "testing"
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
 
 func TestSerial(t *testing.T) {
+	loc, err := time.LoadLocation("Europe/Paris")
```
nineinchnick, line 25:
```suggestion
	paris, err := time.LoadLocation("Europe/Paris")
```

trino/trino_test.go
```diff
@@ -765,6 +765,123 @@ func TestQueryColumns(t *testing.T) {
...
+		actualTypes[i].precision, actualTypes[i].scale, actualTypes[i].hasScale = column.DecimalSize()
+		actualTypes[i].length, actualTypes[i].hasLength = column.Length()
+		actualTypes[i].scanType = column.ScanType()
+	}
+
+	assert.Equal(t, actualTypes, expectedTypes)
+
+	if shouldError {
+		assert.False(t, rows.Next())
+		require.Error(t, rows.Err())
```
nineinchnick, line 877: It would be more readable as two separate tests and you should test for a specific error message, not just the fact that any error is present. If you'd do that, test cases in one of the tests would be simpler as it would not have to check for precision.

nineinchnick (commented):
Just a few nitpicks, almost there!

trino/serial_test.go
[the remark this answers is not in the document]
nineinchnick: I can't resolve comments in PRs in this repository, feel free to mark them as resolved yourself (I think you should be able to).

README.md
```diff
@@ -215,7 +220,8 @@ SELECT * FROM table WHERE col_double = cast(? AS DOUBLE) OR col_timestamp = CAST
 ### Response rows
 
 When reading response rows, the driver supports most Trino data types, except:
-* time and timestamps with precision - all time types are returned as `time.Time`
+* time and timestamps with precision - all time types are returned as `time.Time`.
+  All precisions up to nanoseconds (`TIMESTAMP(9)` or `TIME(9)`) are supported.
```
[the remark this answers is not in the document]
nineinchnick: That can be a bit out of the scope of this PR, since what I requested should go into the description of `trino.NullTime`, in the [Response rows](https://github.com/trinodb/trino-go-client#response-rows) section, but there's not much there now. This should not block this PR from getting merged.

trino/trino_test.go
JPMoresmau: We can still perfectly get the precision for the columns query even though retrieving the actual rows errors. Also the error message is very technical, and not too helpful because it shows the parsing error for the last layout we tried. Should have a more legible error like "value XXX cannot be parsed as a date or time"? Or collect the various errors into one?

nineinchnick: > We can still perfectly get the precision for the columns query even though retrieving the actual rows errors.

This is an even better reason to split these up (or just have a separate test for errors).

> Should have a more legible error like "value XXX cannot be parsed as a date or time"

That would be ideal, but if you think it's out of the scope of this PR we can address it later. Is it still possible to test for a specific error message, even if it's a complex one? I mean can we keep the tests simple?

JPMoresmau: Yes, I've changed the test that errors to use a time literal so we can test for a consistent error message since the order of layouts is deterministic. Maybe a better error message could be done in a separate PR, yes.

JPMoresmau:
@nineinchnick I think I've merged your suggestions and addressed your comments, let me know. I suppose i can squash the commits once you're fine with my latest changes...

trino/serial.go
```diff
@@ -35,6 +35,57 @@ func (e UnsupportedArgError) Error() string {
...
+		location = time.UTC
+	}
+	return trinoTimeTz(time.Date(0, 0, 0, hour, minute, second, nanosecond, location))
+}
+
+// Timestamp indicates we want a TimeStamp type WITHOUT a time zone in Trino from a Golang time.
+type trinoTimestamp time.Time
+
+func StripTimezone(time time.Time) trinoTimestamp {
+	return trinoTimestamp(time)
```
nineinchnick, line 86: You've applied my suggestion to change this in the README but not here.
```suggestion
// Timestamp creates a representation of a Trino Timestamp(9) type.
func Timestamp(year int,
    month time.Month,
    day int,
    hour int,
	minute int,
	second int,
	nanosecond int) trinoTimestamp {
	return trinoTimestamp(time.Date(day, month, year, hour, minute, second, nanosecond, time.UTC)
```

trino/serial.go
```diff
@@ -35,6 +35,57 @@ func (e UnsupportedArgError) Error() string {
...
+	month time.Month
+	day   int
+}
+
+// Date creates a representation of a Trino Date type.
+func Date(year int, month time.Month, day int) trinoDate {
+	return trinoDate{year, month, day}
+}
+
+// Time indicates we want a Time type in Trino from a Golang time.
```
nineinchnick, line 50:
```suggestion
// trinoTime represents a Time type in Trino.
```

trino/serial.go
```diff
@@ -35,6 +35,57 @@ func (e UnsupportedArgError) Error() string {
...
+
+// Time creates a representation of a Trino Time type. To represent time with precision higher than nanoseconds, pass the value as a string and use a cast in the query.
+func Time(hour int,
+	minute int,
+	second int,
+	nanosecond int) trinoTime {
+	return trinoTime{hour, minute, second, nanosecond}
+}
+
+// TimeTz indicates we want a Time With Timezone type in Trino from a Golang time.
```
nineinchnick, line 66:
```suggestion
// trinoTimeTz represents a Time(9) With Timezone type in Trino.
```

trino/serial.go
```diff
@@ -35,6 +35,57 @@ func (e UnsupportedArgError) Error() string {
...
+type trinoTimeTz time.Time
+
+func TimeTz(hour int,
```
nineinchnick, line 69:
```suggestion
// TimeTz creates a representation of a Trino Time(9) With Timezone type.
func TimeTz(hour int,
```

trino/trino_test.go
```diff
@@ -765,6 +765,321 @@ func TestQueryColumns(t *testing.T) {
...
+	}
+
+	assert.Equal(t, actualTypes, expectedTypes)
+
+	assert.True(t, rows.Next())
+	require.NoError(t, rows.Err())
+
+}
+
+func testPrecisionDateTime(t *testing.T, precision int64, shouldError bool) {
```
nineinchnick, line 974: This is now unused, remove it.

nineinchnick (commented):
Thanks for bearing with me. This is the last round from me. LGTM % these nitpicks.

I'm not a maintainer, so we need @hashhar and/or @losipiuk to do the review now. You could squash commits to make reviewing easier for them.

README.md
```diff
@@ -215,7 +220,11 @@ SELECT * FROM table WHERE col_double = cast(? AS DOUBLE) OR col_timestamp = CAST
 ### Response rows
 
 When reading response rows, the driver supports most Trino data types, except:
-* time and timestamps with precision - all time types are returned as `time.Time`
+* time and timestamps with precision - all time types are returned as `time.Time`.
+  All precisions up to nanoseconds (`TIMESTAMP(9)` or `TIME(9)`) are supported (since
```
losipiuk, line 224: why exactly are we constraining ourselve to nanosecond precision? Would it be possible to support up to `TIME*(12)`?

nineinchnick: Because Go's `time.Time` only supports nanoseconds. To support picoseconds, we'd have to define a custom data type. I'm not sure how much work it would be to make it compatible with `time.Time`.

nineinchnick: Recent work in the Python client in https://github.com/trinodb/trino-python-client/pull/300 is also limited but to microseconds.

trino/serial.go
```diff
@@ -35,6 +35,65 @@ func (e UnsupportedArgError) Error() string {
...
+
+// Time creates a representation of a Trino Time type. To represent time with precision higher than nanoseconds, pass the value as a string and use a cast in the query.
+func Time(hour int,
+	minute int,
+	second int,
+	nanosecond int) trinoTime {
+	return trinoTime{hour, minute, second, nanosecond}
+}
+
+// trinoTimeTz represents a Time(9) With Timezone type in Trino.
```
losipiuk, line 66: nit:
```
trinoTimeTz represents a Time With Timezone type in Trino with precision up to 9
```

trino/serial.go
```diff
@@ -35,6 +35,65 @@ func (e UnsupportedArgError) Error() string {
...
+type trinoTimeTz time.Time
+
+// TimeTz creates a representation of a Trino Time(9) With Timezone type.
```
losipiuk, line 69: nit: like above

trino/serial.go
```diff
@@ -35,6 +35,65 @@ func (e UnsupportedArgError) Error() string {
...
+	if location == nil {
+		location = time.UTC
+	}
+	return trinoTimeTz(time.Date(0, 0, 0, hour, minute, second, nanosecond, location))
+}
+
+// Timestamp indicates we want a TimeStamp type WITHOUT a time zone in Trino from a Golang time.
+type trinoTimestamp time.Time
+
+// Timestamp creates a representation of a Trino Timestamp(9) type.
```
losipiuk, line 86: those comments are not really useful? @nineinchnick do we have check which enforces those?

nineinchnick: We don't currently have a linter in the CI that would check for this. The built-in `go vet` doesn't, we'd have to use `staticcheck` (recommended after `go lint` was deprecated) or `golangci-lint` (which is huge but popular).

Having comments for all exported functions is idiomatic in Go and it would show up in online docs like here: https://pkg.go.dev/github.com/trinodb/trino-go-client@v0.308.0/trino

What would it take to make them more useful? Should it explain why `time.Time` is not used instead?

JPMoresmau:
Sorry, is anything more needed from me to get that PR approved?

trino/serial.go
losipiuk: > What would it take to make them more useful

Not sure. Presonally I would just drop it. But if we want to follow Go guildelines let's go with it :)

PR #70, opened 2023-03-02
Update dependencies

### LINKED ISSUE #69: Update Go version to 1.18
- Currently, the project is based on the 1.14 Go version. Let's upgrade the Go version to 1.18
- Also, `golang.org/x/net v0.0.0-20220706163947-c90051bbdb60` package has the [vulnerability ](https://pkg.go.dev/vuln/GO-2023-1571)

nineinchnick:
I ran:
```bash
go get -u ./...
go mod tidy
```

I also updated go versions in GHA workflows.

Go 1.20 changed how `time.Parse()` works with sub-nanosecond precision, so I had to remove one test. I bumped minimum required Go version to 1.17 because of some dependencies. I also added testing with older Trino versions and updated the README. The oldest Trino version we can test is 372 because this is when we added health checks to the Docker image and tests rely on this when starting up. I don't want to modify them.

Fixes #69


nineinchnick:
@losipiuk PTAL, we might want to cut a new release after this, since #69 reported that some dependencies have known vulnerabilities.

losipiuk (approved):

nineinchnick:
Tests are failing because of a change in Go 1.20, from [the release notes](https://go.dev/doc/go1.20) of the `time` package:
> [Parse](https://go.dev/pkg/time/#Parse) now ignores sub-nanosecond precision in its input, instead of reporting those digits as an error.

I'll remove the test and update the README to reflect what happens in that case.

nineinchnick:
@losipiuk all green, PTAL again. I updated the description as I had to raise minimum supported go and trino versions.

PR #71, opened 2023-03-04
Use correct Go version for doing releases

nineinchnick:
In YAML `1.20` gets parsed as a `1.2` float so it needs to be quoted. Also, the `--rm-dist` flag in goreleaser is deprecated and was renamed to `--clean`.


nineinchnick:
@losipiuk can you merge this and retry the release of 0.310? I think we could recreate the tag if needed, there are no code changes.

losipiuk:
Hopefully will be ok nowl.

PR #73, opened 2023-04-12
Add support to pass ssl certificate value instead of only path

mrveera:


trino/trino.go
```diff
@@ -290,11 +296,17 @@ func newConn(dsn string) (*Conn, error) {
...
 		}
-	} else if certPath := query.Get(SSLCertPathConfig); certPath != "" && serverURL.Scheme == "https" {
-		cert, err := ioutil.ReadFile(certPath)
-		if err != nil {
-			return nil, fmt.Errorf("trino: Error loading SSL Cert File: %v", err)
+	} else if serverURL.Scheme == "https" {
+
+		cert := []byte(query.Get(SSLCertConfig))
+
+		if certPath := query.Get(SSLCertPathConfig); certPath != "" {
```
nineinchnick, line 303: It should not be possible to specify both cert and cert path, can you add a condition to check it and return an actionable error message? And a test.

trino/trino_test.go
```diff
@@ -62,6 +62,21 @@ func TestConfigSSLCertPath(t *testing.T) {
 	assert.Equal(t, want, dsn)
 }
 
+func TestConfigSSLCert(t *testing.T) {
+	c := &Config{
+		ServerURI:         "https://foobar@localhost:8080",
+		SessionProperties: map[string]string{"query_priority": "1"},
+		SSLCert:           "cert.pem",
```
nineinchnick, line 69: `SSLCert` should have contents of a cert, it's not a file path like `SSLCertPath`. This is important to test to make sure that a large value with equal signs (base64 encoded) and newlines won't cause any issues here.

trino/trino.go
mrveera: Sure, i will add condition for this

trino/trino_test.go
nineinchnick: @mrveera are you going to continue working on this PR? If not, I can handle this.

nineinchnick: I opened https://github.com/trinodb/trino-go-client/pull/77 with comments addressed.

PR #75, opened 2023-04-12
Support bytes

chein-huang:
translate bytes to 'VARBINARY' in trino so we can filter column which type is ‘VARBINARY’ by bytes


.vscode/launch.json
```diff
@@ -0,0 +1,18 @@
+{
+    // Use IntelliSense to learn about possible attributes.
```
nineinchnick, line 2: Please remove this file

trino/serial.go
```diff
@@ -147,9 +148,8 @@ func Serial(v interface{}) (string, error) {
 	case string:
 		return "'" + strings.Replace(x, "'", "''", -1) + "'", nil
 
-		// TODO - []byte should probably be matched to 'VARBINARY' in trino
 	case []byte:
-		return "", UnsupportedArgError{"[]byte"}
+		return "X'" + strings.ToUpper(hex.EncodeToString(x)) + "'", nil
```
nineinchnick, line 152: This adds support for using `[]byte` in query params, right? What about reading VARBINARY columns in query results?

Also, please update the `README`, there's a section there with a list of supported and not supported data types.

.vscode/launch.json
chein-huang: sry, this file removed now

trino/serial.go
chein-huang: reading VARBINARY columns in query results we need to implement sql.Scanner interface. 
like 
`type TrinoVARBINARY []byte

func (h *TrinoVARBINARY) Scan(src any) (err error) {
	switch s := src.(type) {
	case string:
		var bytes []byte
		bytes, err = base64.StdEncoding.DecodeString(s)
		if err == nil {
			*h = bytes
		}
	default:
		err = fmt.Errorf("unsupported type: %T", src)
	}
	return
}
`

nineinchnick: Would to like to try in this PR? I know I'm asking to extend the scope of work, so it's also ok to do this in a follow up PR

chein-huang: I will do this in next PR. I hope to also support driver.Valuer interface

README.md
```diff
@@ -196,6 +196,7 @@ When passing arguments to queries, the driver supports the following Go data typ
 * integers
 * `bool`
 * `string`
+* `bytes`
```
nineinchnick, line 199: Bytes or []byte?

README.md
```diff
@@ -255,7 +256,10 @@ will receive a `[]interface{}` slice, with values of the following types:
 * `json.Number` for any numeric Trino types
 * `[]interface{}` for Trino arrays
 * `map[string]interface{}` for Trino maps
-* `string` for other Trino types, as character, date, time, or timestamp
+* `string` for other Trino types, as character, date, time, varbinary or timestamp
```
nineinchnick, line 259: This section is about reading data from Trino. We discussed this in another comment that support for []byte will be added in a separate PR, so remove this for now

nineinchnick (approved):

README.md
chein-huang: oh yes, it removed now

nineinchnick:
@cla-bot check

PR #76, opened 2023-04-13
Support read varbinary

chein-huang:
Support read trino type "varbinary" as []byte

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ A [Trino](https://trino.io) client for the [Go](https://golang.org) programming 
   * `float64`, `sql.NullFloat64`
   * `map`, `trino.NullMap`
   * `time.Time`, `trino.NullTime`
+  * `[]byte`, `trino.NullVarbinary`
   * Up to 3-dimensional arrays to Go slices, of any supported type
 
 ## Requirements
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -204,6 +205,8 @@ When passing arguments to queries, the driver supports the following Go data typ
 * the result of `trino.TimeTz(hour, minute, second, nanosecond, location)` - passed to Trino as a time with a time zone
 * the result of `trino.Timestamp(year, month, day, hour, minute, second, nanosecond)` - passed to Trino as a timestamp without a time zone
 
+We can also implement driver.Valuer interface to return data of types above.
```
sjones94549, line 208:
```suggestion
We can also implement the `driver.Valuer` interface to return data of the types above.
```

README.md
```diff
@@ -246,6 +249,7 @@ To read query results containing arrays or maps, pass one of the following struc
 * `trino.NullSliceFloat64`
 * `trino.NullSliceTime`
 * `trino.NullSliceMap`
+* `trino.NullSliceByte`
```
nineinchnick, line 252: Adding `trino.NullSliceByte` here might be misleading, as it's used for `VARBINARY`. This section is about arrays and maps, so `ARRAY(VARBINARY)` will use `trino.NullSlice2Byte`.

nineinchnick: Add `VARBINARY` to the list in the beginning of this section, maybe after `UUID`.

trino/trino.go
```diff
@@ -2185,6 +2200,169 @@ func (s *NullSlice3Map) Scan(value interface{}) error {
...
+// Scan implements the sql.Scanner interface.
+func (b *NullSliceByte) Scan(value interface{}) error {
+	if value == nil {
+		b.SliceByte, b.Valid = []byte{}, false
+		return nil
+	}
+	b.Valid = true
+
+	var err error
+	switch v := value.(type) {
```
nineinchnick, line 2232: Can the server really return `[]byte` here or is it always a base64 encoded string? If the latter, remove the switch statement.

PR #77, opened 2023-05-17
Add support to pass ssl certificate as a string, not a file path

nineinchnick:
Supersedes #73, applied review comments.

This is required to use this driver in https://github.com/starburstdata/grafana-trino/ and get that plugin certified by Grafana. Grafana now does not allow plugins to make any file system calls, so we need a way to pass a certificate as a value, instead of saving it to a temp file.

### DIFF
--- a/trino/trino.go
+++ b/trino/trino.go
@@ -203,11 +203,28 @@ func (c *Config) FormatDSN() (string, error) {
 	KerberosEnabled, _ := strconv.ParseBool(c.KerberosEnabled)
 	isSSL := serverURL.Scheme == "https"
 
-	if isSSL && c.SSLCertPath != "" {
+	if c.CustomClientName != "" {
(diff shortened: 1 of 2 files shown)


nineinchnick:
@losipiuk PTAL

trino/trino.go
```diff
@@ -203,11 +203,28 @@ func (c *Config) FormatDSN() (string, error) {
 	KerberosEnabled, _ := strconv.ParseBool(c.KerberosEnabled)
 	isSSL := serverURL.Scheme == "https"
 
-	if isSSL && c.SSLCertPath != "" {
+	if c.CustomClientName != "" {
```
losipiuk, line 206: Oh - cool -next commit :)

losipiuk: nit: can we test mutex logic maybe?

losipiuk (approved):

PR #85, opened 2023-08-25
Update dependencies

nineinchnick:
This will also allow testing with latest Go (1.21) which came out after last release of this driver, and with latest Trino (425). We don't have to bump these versions anywhere, as we're using the `latest` tag in the CI.


nineinchnick:
Correction, `trinodb/trino:latest` still points to 424 as there was some issue with releasing the 425 Docker image.

losipiuk (approved):

PR #90, opened 2023-10-08
Support reading float literals returned as strings

nineinchnick:
Trino 428 removed the deprecated `parse_decimal_literals_as_double` property. Without this, the server to returns decimal literals as strings, not floats. Support reading such values when scanning results.

Ref: https://github.com/trinodb/trino/pull/19166


losipiuk (approved):

losipiuk:
thanks

PR #91, opened 2023-10-11
feat: support custom request headers

patrickdemers6:
Depending on the Trino server configuration, the ability to pass custom headers on a per-request is incredibly valuable. Transaction ID, instrumentation ID, and Authorization headers to name a few. This adds the ability to set custom headers using a `sql.Named` parameter where the name has the prefix `X-Header-`.

# Testing
1. Unit test in `trino_test.go`.
2. Manually validated with a local Trino server.

I signed CLA earlier today but it may not be in the system yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,17 @@ db.Query("SELECT * FROM foobar WHERE id=?", 1, sql.Named("X-Trino-User", string(
 
 The position of the X-Trino-User NamedArg is irrelevant and does not affect the query in any way.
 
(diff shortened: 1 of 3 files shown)


nineinchnick:
There's a similar discussion in the main Trino repository: https://github.com/trinodb/trino/pull/15826. I think it also applies here. I'd rather discuss specific use cases and keep Trino clients in multiple languages consistent.

PR #92, opened 2023-10-17
Use http.NewRequestWithContext

devinsba:
Specifically to support tracing we need to pass the existing context into the downstream http requests

### DIFF
--- a/trino/trino.go
+++ b/trino/trino.go
@@ -448,8 +448,8 @@ func (c *Conn) Close() error {
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


devinsba:
`TestIntegrationQueryContextCancellation` is failing with these changes so I'll have to look into that next week. I'm fairly new to go contexts though so if someone has any helpful hints for failures [here](https://github.com/trinodb/trino-go-client/blob/master/trino/integration_test.go#L703) I'd appreciate it

nineinchnick:
@devinsba my attempt is in #95. I figured out that when cancelling the query by sending a DELETE request, we need to "uncancel" the parent context using `context.WithoutCancel()`.

Issue #34, opened 2022-01-04
Add tests for UPDATE and DELETE


hashhar:
From https://github.com/trinodb/trino-go-client/issues/32#issuecomment-1003227977

PR #93, opened 2023-10-19
Support nil as query parameters

### LINKED ISSUE #32: plan to support Update/Insert/Delete ?
Hi, I found the README file said that: 
_Only read operations are supported, such as SHOW and SELECT._

Is there any plan to support Update/Insert/Delete? 

Thanks in advance.

### LINKED ISSUE #84: Null support for inserts
Client version: v0.312.0

I have a use case where I insert multiple records with different fields are being nullable. For example, my table schema is as follows:

```
id: int
name: string
order: int (nullable)
url: string (nullable)
```

`order` and `url` fields are nullable so when I have 2 records:

(1, "first record", nil, "url")
(2, "second record", 1, nil)

I expected to execute the following SQL query (with arguments):

```
INSERT INTO my_table (id, name, order, url)
VALUES
(1, 'first record', NULL, 'url'),
(2, 'second record', 1, NULL);
```

However, when `ExecContext` function is called with `driver.NamedValue` that has `nil` as a value, it throws exception in type conversion function [here](https://github.com/trinodb/trino-go-client/blob/7593b22e2d5af23ace8966fd3bf8dbe08d59f91a/trino/serial.go#L101-L102).

To support nil values, we can change the type conversion for nil to:

```
case nil:
    return "NULL", nil
```

I am curious what your thoughts are. If this is not a desirable behaviour, is there another workaround for the problem?

nineinchnick:
Fixes #84, fixes #32, but not #34 because the memory connector doesn't support UPDATE and DELETE queries and I didn't want to start another container.

### DIFF
--- a/trino/trino_test.go
+++ b/trino/trino_test.go
@@ -1739,10 +1739,13 @@ func TestExec(t *testing.T) {
 		assert.NoError(t, db.Close())
 	})
 
-	result, err := db.Exec("CREATE TABLE memory.default.test (id INTEGER, name VARCHAR)")
+	result, err := db.Exec("CREATE TABLE memory.default.test (id INTEGER, name VARCHAR, optional VARCHAR)")
 	require.NoError(t, err, "Failed executing CREATE TABLE query")
 
-	result, err = db.Exec("INSERT INTO memory.default.test (id, name) VALUES (?, ?), (?, ?), (?, ?)", 123, "abc", 456, "def", 789, "ghi")
+	result, err = db.Exec("INSERT INTO memory.default.test (id, name, optional) VALUES (?, ?, ?), (?, ?, ?), (?, ?, ?)",
+		123, "abc", nil,
+		456, "def", "present",
+		789, "ghi", nil)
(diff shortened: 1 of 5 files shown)


trino/trino_test.go
```diff
@@ -1719,3 +1719,52 @@ func BenchmarkQuery(b *testing.B) {
 		rows.Close()
 	}
 }
+
+func TestExec(t *testing.T) {
```
losipiuk, line 1723: I would argue this does not test INSERT (what it does per commit title) - give no support of that in connector.

nineinchnick: Sorry, I don't understand. The memory connector supports inserts but doesn't support updates and deletes. And this tests executes an `INSERT` query, and passes.

losipiuk (approved):

losipiuk:
wanna release with this one?

nineinchnick:
Yes, let's do a release.

losipiuk:
release workflow: https://github.com/trinodb/trino-go-client/actions/runs/6586703072

PR #95, opened 2023-10-22
Add OpenTelemetry instrumentation

nineinchnick:
Pass parent context to HTTP requests and use the OTEL transport to make sure that traces are propagated. The OTEL transport will also create spans for all HTTP requests.

This matches the current behavior of the Trino JDBC driver.

Supersedes #92 

I'll keep this as a draft before doing some manual testing.


nineinchnick:
@wendigo @losipiuk the OTEL libraries require Go 1.20. Do you think we can raise the minimum required Go version here? Go 1.20 was released Feb. 1, 2023. On the other hand, there are very few new features in this driver, so users who can't immediately upgrade Go can lock in an older version of this driver.

trino/trino.go
```diff
@@ -307,7 +308,9 @@ func newConn(dsn string) (*Conn, error) {
 		}
 	}
 
-	var httpClient = http.DefaultClient
+	var httpClient = &http.Client{
+		Transport: otelhttp.NewTransport(http.DefaultTransport),
```
nineinchnick, line 312: If users can pass a custom http client, we don't really need to use the OTEL transport by default.

PR #96, opened 2023-10-23
Allow OpenTelemetry instrumentation

nineinchnick:
Pass parent context to HTTP requests and document how to use the OTEL transport to make sure that traces are propagated. The OTEL transport will also create spans for all HTTP requests.

An alternative to #95.

Supersedes #92


trino/trino.go
```diff
@@ -983,12 +988,12 @@ func (qr *driverRows) Close() error {
 	if qr.stmt.user != "" {
 		hs.Add(trinoUserHeader, qr.stmt.user)
 	}
-	req, err := qr.stmt.conn.newRequest("DELETE", qr.stmt.conn.baseURL+"/v1/query/"+url.PathEscape(qr.queryID), nil, hs)
+	ctx, cancel := context.WithTimeout(context.WithoutCancel(qr.ctx), DefaultCancelQueryTimeout)
```
nineinchnick, line 991: Whoops, `context.WithoutCancel()` was added in Go 1.21. To keep supporting older versions (>=1.17), we have to either:
1. Create a wrapper, like suggested here: https://stackoverflow.com/a/54132324
2. Create a new context and try to copy some known values, like OTEL spans and/or traces

Option 2 sounds safer. I now started wondering if `context.WithoutCancel()` would actually work with `context.WithTimeout()`.

devinsba: 3 sounds good for my use as long as the copy mechanism is configurable, either the function used, or the ability to add keys that will be copied

nineinchnick: It would have to be similar to how the custom http client is injected, which is a bit cumbersome. I'll try to write a test for option 2.

nineinchnick: Since a few months passed already, I'd go back to option 1 - update minimum required Go version. Users can keep using the current driver version until they can migrate to latest Go.

losipiuk (approved):
LGTM % I did not read the docs

nineinchnick:
The docs include an example how to use OpenTelemetry, so it's important. Ignore the other commit that formats the README.

README.md
```diff
@@ -209,6 +209,19 @@ trino.RegisterCustomClient("foobar", foobarClient)
...
+A custom client can also be used to add OpenTelemetry instrumentation. The
+[otelhttp](https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp)
+package provides a transport wrapper that creates spans for HTTP requests and
+propagates the trace ID in HTTP headers:
+
+```go
+otelClient := &http.Client{
+    Transport: otelhttp.NewTransport(http.DefaultTransport),
+}
+trino.RegisterCustomClient("otel", otelClient)
```
losipiuk, line 221: this does not return err?

losipiuk (approved):

README.md
nineinchnick: It does, but the example above didn't check for it. I don't think error handling must be in every example, it would make it harder to read and since this is users code, we don't know a good way to handle errors anyway. Also, common Go linters detect unhandled errors.

Issue #97, opened 2023-10-26
Expose underlying trino error codes / names


deigote:
ATM the only visible attribute on error is the http status code (from `ErrQueryFailed`), but a client might want to react differently depending on the underlying trino error (for example, if it's a exceeded time limit the error might bubble up, whereas if it's an exceeded cpu limit it could back off and retry). Right now the information is avail in the `stmtError`, but that's not exported.

chenzeshenga:
same feature required, thanks in advance

ornew:
I agree with this issue.

However, I don't think #98 is a fundamental solution. To capture a broader range of use cases, I believe we should provide a way for users to obtain details about the errors.

The first issue is that `stmtError` is not exported and cannot be `Unwrap`. When a `stmtError` occurs, it is set as the `Reason` for `ErrQueryFailed`, but since it cannot be  `Unwrap`, users are unable to retrieve the details. `ErrQueryFailed` should be made  `Unwrap`-able, and `stmtError` should be exported.  `Unwrap` would also allow for the identification of errors during transport layers.

Another issue is the omission of fields that may be necessary for precise problem identification. `stmtError` and `stmtFailureInfo` may include stack information and multiple causes. We should define the missing fields and ensure detailed information is decoded.

These changes would allow for the propagation of information to users through the error values and make it possible to identify errors. Adding information to error messages would be also enabled by these changes.

PR #98, opened 2023-10-26
Include error code and name on error message

deigote:
Ideally we'd expose the error info in a structured way like suggested in https://github.com/trinodb/trino-go-client/issues/97 . Since that's more involved as requires changing exported interfaces, in the meantime having the info in the error message might be already helpful for telemetry.

### DIFF
--- a/trino/trino.go
+++ b/trino/trino.go
@@ -725,7 +725,7 @@ type stmtErrorFailureInfo struct {
 }
 
 func (e stmtError) Error() string {
-	return e.FailureInfo.Type + ": " + e.Message
+	return fmt.Sprintf("%v (name: %s, code: %d): %s", e.FailureInfo.Type, e.ErrorName, e.ErrorCode, e.Message)
 }
 
 type stmtStage struct {


trino/trino.go
```diff
@@ -725,7 +725,7 @@ type stmtErrorFailureInfo struct {
 }
 
 func (e stmtError) Error() string {
-	return e.FailureInfo.Type + ": " + e.Message
+	return fmt.Sprintf("%v (name: %s, code: %d): %s", e.FailureInfo.Type, e.ErrorName, e.ErrorCode, e.Message)
```
nineinchnick, line 728: Can you add a simple test for that? Maybe a query with a syntax error? I'd like to see some example value and make sure it's human-readable and informative.

We should not encourage parsing this string and instead fix the interfaces as suggested in #97.

deigote: Can do, but wouldn't a test encourage folk to rely on the message as if it was some sort of contract 🤔 ?

nineinchnick: PTAL at #109, maybe it solves getting error details without having to parse the error string.

nineinchnick:
@deigote I'll close this since we got #97 merged. Feel free to reopen this or create a new issue.

PR #109, opened 2024-04-14
Export Trino errors with details

### LINKED ISSUE #97: Expose underlying trino error codes / names

nineinchnick:
Fixes #97

### DIFF
--- a/trino/trino.go
+++ b/trino/trino.go
@@ -560,6 +560,11 @@ func (e *ErrQueryFailed) Error() string {
 		e.StatusCode, http.StatusText(e.StatusCode), e.Reason)
 }
 
+// Unwrap implements the unwrap interface.
+func (e *ErrQueryFailed) Unwrap() error {
+	return e.Reason
+}
+
 func newErrQueryFailedFromResponse(resp *http.Response) *ErrQueryFailed {
 	const maxBytes = 8 * 1024
 	defer resp.Body.Close()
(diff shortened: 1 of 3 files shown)


trino/trino.go
```diff
@@ -712,27 +717,43 @@ type stmtStats struct {
...
+	Type          string        `json:"type"`
+	Message       string        `json:"message"`
+	Cause         *FailureInfo  `json:"cause"`
+	Suppressed    []FailureInfo `json:"suppressed"`
+	Stack         []string      `json:"stack"`
+	ErrorInfo     ErrorInfo     `json:"errorInfo"`
+	ErrorLocation ErrorLocation `json:"errorLocation"`
+}
+
+func (i ErrTrino) Error() string {
```
wendigo, line 745: move it close to ErrTrino type definition

trino/trino.go
```diff
@@ -712,27 +717,43 @@ type stmtStats struct {
...
+	SqlState      string        `json:"sqlState"`
+	ErrorCode     int           `json:"errorCode"`
+	ErrorName     string        `json:"errorName"`
+	ErrorType     string        `json:"errorType"`
+	ErrorLocation ErrorLocation `json:"errorLocation"`
+	FailureInfo   FailureInfo   `json:"failureInfo"`
 }
 
-type stmtErrorLocation struct {
+type ErrorLocation struct {
```
wendigo, line 730: Move it above the ErrTrino along with FailureInfo

wendigo (approved):

trino/trino.go
nineinchnick: It reads better if type definitions are placed after their first usage.

losipiuk (approved):

PR #112, opened 2024-04-25
support Geometry and Geography values as WKT strings

fengctor:
Allows us to use the WKT strings returned from Geometry and SphericalGeography expressions in queries like `select st_point(0, 0)` and `select to_spherical_geography(st_point(0, 0))` (both giving `"Point (0 0)"`). This would previously result in a `type not supported: "SphericalGeography"` error.

### DIFF
--- a/trino/trino.go
+++ b/trino/trino.go
@@ -1469,7 +1469,7 @@ func (c *typeConverter) ConvertValue(v interface{}) (driver.Value, error) {
 			return nil, err
 		}
 		return vv.Bool, err
-	case "json", "char", "varchar", "varbinary", "interval year to month", "interval day to second", "decimal", "ipaddress", "uuid", "unknown":
+	case "json", "char", "varchar", "varbinary", "interval year to month", "interval day to second", "decimal", "ipaddress", "uuid", "SphericalGeography", "unknown":
 		vv, err := scanNullString(v)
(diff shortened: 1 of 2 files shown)


trino/trino.go
```diff
@@ -1469,7 +1469,7 @@ func (c *typeConverter) ConvertValue(v interface{}) (driver.Value, error) {
 			return nil, err
 		}
 		return vv.Bool, err
-	case "json", "char", "varchar", "varbinary", "interval year to month", "interval day to second", "decimal", "ipaddress", "uuid", "unknown":
+	case "json", "char", "varchar", "varbinary", "interval year to month", "interval day to second", "decimal", "ipaddress", "uuid", "SphericalGeography", "unknown":
```
nineinchnick, line 1472: Should we also add `Geometry` here?

nineinchnick (approved):

trino/trino.go
fengctor: Yup, I can also add that

fengctor:
Hi @nineinchnick, it looks like I need help from someone with write access to get this merged.

nineinchnick:
I'm not a maintainer here, @wendigo @losipiuk PTAL

wendigo (approved):

PR #115, opened 2024-04-30
Use latest Go for releases

nineinchnick:
#96 bumped the minimum required version to 1.21, so running releases with 1.20 will likely fail.

Also run the CI workflow only on pushes to the master branch. Now when working on PRs from forks, it runs both in the target repo and in the fork, which is unnecessary.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,8 +1,10 @@
 name: ci
 
 on:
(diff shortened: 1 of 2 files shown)


wendigo (approved):

wendigo:
@nineinchnick ping me when CI is green

PR #117, opened 2024-05-03
Don't fail when decoding empty strings as floats in server responses

### LINKED ISSUE #116: Queries sometimes fail when decoding progress percent in the response
As reported in https://github.com/trinodb/grafana-trino/issues/258, queries sometimes fail with:
```
error querying the database: trino: json: cannot unmarshal string into Go struct field stmtStats.stats.progressPercentage of type float32
```

The `progressPercent` field is declared as a `float32`: https://github.com/trinodb/trino-go-client/blob/master/trino/trino.go#L722, but the Trino server uses an `OptionalDouble`: https://github.com/trinodb/trino/blob/ca209630136eabda2449594ef2b6a4d82fb9c2e5/core/trino-main/src/main/java/io/trino/server/QueryProgressStats.java#L132

The driver should ignore invalid values in stats.

nineinchnick:
Fixes #116 

I'm still not sure if the root cause is an empty string, that's why this is a draft.

### DIFF
--- a/trino/trino.go
+++ b/trino/trino.go
@@ -713,27 +713,27 @@ type stmtResponse struct {
 }
 
 type stmtStats struct {
-	State                string    `json:"state"`
-	Scheduled            bool      `json:"scheduled"`
-	Nodes                int       `json:"nodes"`
-	TotalSplits          int       `json:"totalSplits"`
-	QueuesSplits         int       `json:"queuedSplits"`
-	RunningSplits        int       `json:"runningSplits"`
-	CompletedSplits      int       `json:"completedSplits"`
-	UserTimeMillis       int       `json:"userTimeMillis"`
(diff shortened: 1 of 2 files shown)


eladitzhakian:
This is a major issue in a project I'm working on ☝️

nineinchnick:
@eladitzhakian can you help with providing some reproduction steps? The test I added here is very synthetic.

eladitzhakian:
@nineinchnick in our use case it happens when trino is OOM while spawning a new query. set `-Xmx` to a relatively low value then execute a query

eladitzhakian:
@nineinchnick any chance to push this for the next release? 🙏

nineinchnick:
I don't think so, the work here stalled. I don't have a good reproduction of the issue.

PR #119, opened 2024-05-23
Parsing of datetime to take UTC if timezone is not provided

silent-lad:


trino/trino.go
```diff
@@ -2029,7 +2029,7 @@ func parseNullTime(v string) (NullTime, error) {
 	var t time.Time
 	var err error
 	for _, layout := range timeLayouts {
-		t, err = time.ParseInLocation(layout, v, time.Local)
+		t, err = time.ParseInLocation(layout, v, time.UTC)
```
nineinchnick, line 2032: Please also update tests.

silent-lad: My bad, ran the tests on a different branch and sent it. Fixing this

silent-lad: This is done

silent-lad:
Have fixed the tests @nineinchnick

nineinchnick:
I've thought about this, and this is a breaking change for anyone using this data type. What's worse, it's silent, meaning the behavior will just change for anyone not pinned to a specific version of this driver. Even if this change is correct, we should only do this in a `v2` version.

If you don't mind, I'd keep this open for a while, hoping we can get more feedback from other people. In the meantime, I'd suggest using a workaround, like you can create a struct implementing `sql.Scanner` in your application and pass it to `Scan()` instead of `trino.Time`.

silent-lad:
Can we have the default timezone passed with the connection string like we have for custom http client

```
if tzKey := query.Get("default_tz"); tzKey != "" {
                 os.Setenv("TZ", default_tz)
}
```

That way its both configurable and backwards compatible

nineinchnick:
@cla-bot check

nineinchnick:
@silent-lad I like your last suggestion, can we get back to this? You haven't signed the CLA yet.

silent-lad:
https://github.com/trinodb/cla/pull/8/files
I have a PR for the CLA. Haven't gotten it approved. @nineinchnick

nineinchnick:
PRs are not processed in that repository, you need to send the signed CLA via email.

silent-lad:
> can we get back to this? 

I will update the PR with changes today. 

> you need to send the signed CLA via email.

Is the procedure documented somewhere so I can follow?

nineinchnick:
> Is the procedure documented somewhere so I can follow?

Yes, please read the `cla-bot` comments.

nineinchnick:
@cla-bot check

PR #120, opened 2024-06-01
Update dependencies

nineinchnick:
This is a minor update, I just want to run the CI with latest Trino, 449.


nineinchnick:
The integration tests use a copy of the `jvm.config` file, which requires `-XX:+UnlockDiagnosticVMOptions -XX:G1NumCollectionsKeepPinned=10000000` in the latest version. I think we should wait for the next Trino release, where the base image will get updated to Java 22.0.2

PR #127, opened 2024-10-06
Escape values of extra credentials and session properties

### LINKED ISSUE #125: Special characters support in extra credentials.
Query fails if there are following special characters in the extra credentials.

= Message: trino: query failed (400 Bad Request): "Invalid X-Trino-Extra-Credential header"

% Message: trino: query failed (400 Bad Request): "Invalid X-Trino-Extra-Credential header: java.lang.IllegalArgumentException: URLDecoder: Illegal hex characters in escape (%) pattern - Error at index 1 in: "do""

+  Message: trino: query failed (200 OK): "io.trino.spi.TrinoException: [10]: authentication failed: Detailed info for this error can be found with correlation ID '6BDD97D810581241A472BA8974EEF6D9'"

, Message: trino: query failed (400 Bad Request): "Invalid X-Trino-Extra-Credential header"

These special characters are commonly used in passwords and should be supported.

According to the comment here https://github.com/trinodb/trino/issues/21751#issuecomment-2110780864 Each client should be responsible for url encoding the values and Java client has already done that.

nineinchnick:
Correctly escape values of extra credentials and session properties when
building HTTP headers.
    
Change the format of extra credentials and session properties in the URL
to match the Trino JDBC and Python drivers - keys and values are
separated with a colon (`:`) and multiple entries are separated with a
semicolon (`;`). Note: all special characters, including the semicolon,
must be URL escaped.

Fixes #125

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,17 @@
+changelog:
+  exclude:
+    labels:
+      - ignore-for-release
+  categories:
+    - title: Breaking changes
+      labels:
+        - breaking-change
+    - title: Features
+      labels:
+        - enhancement
+    - title: Bug fixes
+      labels:
+        - bug
+    - title: Other changes
+      labels:
+        - "*"
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,2 +1,4 @@
 builds:
 - skip: true
+changelog:
+  use: github-native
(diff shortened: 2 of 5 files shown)


nineinchnick:
@wendigo @hashhar PTAL

hashhar (approved):

PR #128, opened 2024-10-06
Support EXECUTE IMMEDIATE

### LINKED ISSUE #118: How to set a custom name to prepared statements, in different connections
I'm trying to name queries differently to avoid the generic
`EXECUTE _trino_go USING 4`
and get some context (at least in names).
`EXECUTE total_customers_transactions USING 4`
Is there a way to achieve this using the lib via ExecContext() o QueryRowContext()?

I feel like I'm missing something.
Thank you in advanced!

### LINKED ISSUE #123: I use QueryContext to query a super long sql (13700 bytes). When there are args parameters, the error http2ErrCodeEnhanceYourCalm (11) will be reported. Has anyone encountered this?
image

 I use QueryContext to query a super long sql (13700 bytes). When there are args parameters, the error http2ErrCodeEnhanceYourCalm (11) will be reported. Has anyone encountered this?

nineinchnick:
Use EXECUTE IMMEDIATE sent in the HTTP request body, instead of putting the query text in HTTP headers. This should allow sending large query text. It can be enabled by setting the `explicitPrepare` option to false in the connection string.

Fixes #123 and fixes #118

### DIFF
--- a/trino/etc/config.properties
+++ b/trino/etc/config.properties
@@ -12,3 +12,5 @@ http-server.https.port=8443
 http-server.authentication.allow-insecure-over-http=true
 http-server.https.keystore.path=/etc/trino/secrets/certificate_with_key.pem
 internal-communication.shared-secret=gotrino
+
+query.max-length=5000043
--- a/trino/etc/jvm.config
+++ b/trino/etc/jvm.config
@@ -1,4 +1,4 @@
--Xmx1G
+-Xmx4G
 -XX:+UseG1GC
 -XX:G1HeapRegionSize=32M
 -XX:+UseGCOverheadLimit
--- a/trino/integration_test.go
+++ b/trino/integration_test.go
@@ -34,6 +34,7 @@ import (
 	"math/big"
 	"net/http"
 	"os"
+	"strconv"
 	"strings"
 	"testing"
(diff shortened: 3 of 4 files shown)


nineinchnick:
Nope, you need to enable `explicitPrepare` to use `EXECUTE IMMEDIATE`, it's off by default, same as in the JDBC driver.

For tracking breaking changes in release notes, I'm adding more configuration in #127

hashhar:
> Nope, you need to enable explicitPrepare to use EXECUTE IMMEDIATE, it's off by default, same as in the JDBC driver.

I see now but the comment is confusing - `useExplicitPrepare = true` by default in TrinoConnection which means we default to using explicit prepare.

hashhar:
> It can be enabled by setting the explicitPrepare option to true in the connection string.

Should be `It can be enabled by setting the explicitPrepare option to false in the connection string.`.

nineinchnick:
I might have messed it up because I initially assumed the JDBC driver would have this turned **on** by default, only to later notice it's the opposite, so I also changed it here accordingly. Will double-check, thanks for noticing!

hashhar:
the implementation seems correct, just the PR description seems wrong/

Issue #60, opened 2022-11-08
Trino VARBINARY columns should scan to []byte slices instead of strings.


ademille:
Trino columns that are VARBINARY type are base64 encoded before being sent via HTTP to the go client.  The current client implementation scans the data into a 'string' type without base64 decoding the values, forcing the user of the driver to do an extra non-intuitive base64 decode to get the original binary data back.

It would be better to have the driver perform the base64 decode and return a []byte slice.
For example, if an IP address or MAC address is stored as a varbinary column in trino, the client could return a []byte type which could be used to create a net.IP or net.HardwareAddr directly.

fatpa:
use `from_utf8({varbinary_data})` function can return base64 decode data

Flgado:
@nineinchnick  this is close right ?

PR #135, opened 2025-02-27
feat: Support `VARBINARY`

hudclark:
# Overview
This PR adds support for encoding/decoding the `[]byte` go type as `VARBINARY` data.

# Additional Information
To keep the scope of this PR small, I did _not_ add support for reading `VARBINARY` values within complex types. This is not a regression, but a gap in the serialization/deserialization code. Timestamps, Numbers, etc also have the same issue - they are read into go types right off of the decoded JSON. This means that a `ROW(X'68656C6C6F')` is deserialized as a `[]interface{}{"aGVsbG8="}`. I did update the README to document this behavior.

To showcase this behavior, I've added a (skipped) failing test case. Given the behavior is documented, changing it would be a breaking change. Perhaps an option could be added in the future to recursively parse complex types such that this test case passes?

I'm happy to drop the commit with the skipped test case if we feel it adds noise or does not represent the desired state.

Resolves:
- https://github.com/trinodb/trino-go-client/issues/60

---

> [!NOTE]
> I have signed and emailed an individual CLA.

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,6 +258,7 @@ types:
 * integers
 * `bool`
 * `string`
+* `[]byte`
 * slices
 * `trino.Numeric` - a string representation of a number
 * `time.Time` - passed to Trino as a timestamp with a time zone
@@ -324,7 +325,11 @@ following types:
 * `json.Number` for any numeric Trino types
 * `[]interface{}` for Trino arrays
 * `map[string]interface{}` for Trino maps
-* `string` for other Trino types, as character, date, time, or timestamp
+* `string` for other Trino types, as character, date, time, or timestamp.
+
+> [!NOTE]
+> `VARBINARY` columns are returned as base64-encoded strings when used within
+> `ROW`, `MAP`, or `ARRAY` values.
 
 ## License
 
--- a/trino/integration_test.go
+++ b/trino/integration_test.go
@@ -727,6 +727,62 @@ func TestIntegrationTypeConversion(t *testing.T) {
 	}
 }
 
+func TestComplexTypes(t *testing.T) {
+	// This test has been created to showcase some issues with parsing
+	// complex types. It is not intended to be a comprehensive test of
+	// the parsing logic, but rather to provide a reference for future
+	// changes to the parsing logic.
+	//
+	// The current implementation of the parsing logic reads the value
+	// in the same format as the JSON response from Trino. This means
+	// that we don't go further to parse values as their structured types.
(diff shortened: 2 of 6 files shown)


trino/integration_test.go
```diff
@@ -110,6 +112,14 @@ func TestMain(m *testing.M) {
...
 					"8443/tcp",
 				},
+			}, func(hc *docker.HostConfig) {
+				hc.Ulimits = []docker.ULimit{
+					{
+						Name: "nofile",
+						Hard: 4096,
+						Soft: 4096,
+					},
+				}
```
hudclark, line 122: Trino requires a minimum of 4096 descriptors in order to startup. On my linux machine, I found docker defaults to a limit of 1024.

I've updated the test configuration such that we can be sure the container is configured with reasonable ulimits.

README.md
```diff
@@ -317,14 +318,18 @@ structs to the `Scan()` function:
...
 To read `ROW` values, implement the `sql.Scanner` interface in a struct. Its
 `Scan()` function receives a `[]interface{}` slice, with values of the
 following types:
 * `bool`
 * `json.Number` for any numeric Trino types
 * `[]interface{}` for Trino arrays
 * `map[string]interface{}` for Trino maps
-* `string` for other Trino types, as character, date, time, or timestamp
+* `string` for other Trino types, as character, date, time, or timestamp.
+  * `VARBINARY` columns are returned as base64-encoded strings
```
nineinchnick, line 332: Is this supposed to be a one item sub-list? It looks weird, maybe add it as a note below the list.

nineinchnick (approved):
Looks great! Once the CLA is cleared, I can merge it. Please remove prefixes from commit messages.

hudclark:
Appreciate the quick review @nineinchnick! I've removed the commit message prefixes and updated the readme to resolve that strange formatting.

I also want to call out that this might be considered a breaking change for clients who have previously been reading in `VARBINARY` as strings, then doing base64 decoding client-side.

hudclark:
I will investigate and fix the failing test shortly

nineinchnick:
@cla-bot check

hudclark:
Hey @nineinchnick, looks like I've been added now https://github.com/trinodb/cla/commit/4053dbd8a21141cb5f9dbd984d0998ce5d754a7e.

Mind trying again? Thanks!

nineinchnick:
Released as https://github.com/trinodb/trino-go-client/releases/tag/v0.322.0

Issue #102, opened 2024-01-11
Do not send DELETE after QueryRow


pajaks:

nineinchnick:
When calling `db.QueryRowContext()` and scanning the result, the `Scan()` method calls `Close()` on the driver's result struct. The driver is not aware that there are no more results to read and issues an additional `DELETE` HTTP request. Because the query no longer exists on the server, it returns a 404 response, which is then returned to the calling code.

The driver should correctly set the internal err to `io.EOF` if there's no data and `nextUri` in the last received response.

chris-wilson-r7:
Any update or workaround for this? Currently I'm having to use Query instead of QueryRow to get around this error?

huangc28:
I bumped to this issue, wow it's an issue known for  a year already. Is this repo still being maintained?

PR #136, opened 2025-03-03
Fix unnecessary cancel requests

Flgado:
# Overview
This PR fixes unnecessary delete requests when the query is already finished.
# Explanation
This change addresses Issue #102 
- When using `QueryRow` followed by a `Scan`, the Trino driver first calls the `Next` method and then the Close method. The `Close` method triggers a delete request to cancel the query, even when the query has already completed.
- My initial approach was to use `nextUri` in the `Close` method instead of directly calling `/v1/query/queryId`, as suggested in the documentation and implemented by other drivers. However, to make this work, I needed to always update `rows.nextUri` inside the `fetch` method, as it was previously retaining only the first URI set [here](https://github.com/Flgado/trino-go-client/blob/bc0714dc52af6f669b36b07f08f1af71e1287c2a/trino/trino.go#L891-L898). Despite these changes, the issue persisted.
- To resolve this, I modified the `Next` method so that when all rows have been read, it fetches the next set of rows. This works because our `fetch` method, along with the running Goroutines, follows `nextUri` until data is retrieved.
  -  For `QueryRow`, this ensures that when the query reaches the finished state with no more data, nextUri is cleared, preventing an unnecessary request on Close().
  -  For `QueryContext`, this is also fine because it fetches the next batch of data, ensuring that additional data (if available) is retrieved correctly.

### DIFF
--- a/trino/trino.go
+++ b/trino/trino.go
@@ -1138,27 +1138,32 @@ func (qr *driverRows) Close() error {
 	if qr.err == sql.ErrNoRows || qr.err == io.EOF {
 		return nil
 	}
+
 	qr.err = io.EOF
 	hs := make(http.Header)
 	if qr.stmt.user != "" {
 		hs.Add(trinoUserHeader, qr.stmt.user)
 	}
-	ctx, cancel := context.WithTimeout(context.WithoutCancel(qr.ctx), DefaultCancelQueryTimeout)
-	defer cancel()
-	req, err := qr.stmt.conn.newRequest(ctx, "DELETE", qr.stmt.conn.baseURL+"/v1/query/"+url.PathEscape(qr.queryID), nil, hs)
-	if err != nil {
-		return err
-	}
-	resp, err := qr.stmt.conn.roundTrip(ctx, req)
-	if err != nil {
-		qferr, ok := err.(*ErrQueryFailed)
-		if ok && qferr.StatusCode == http.StatusNoContent {
-			qr.nextURI = ""
-			return nil
+
+	if qr.nextURI != "" {
+		ctx, cancel := context.WithTimeout(context.WithoutCancel(qr.ctx), DefaultCancelQueryTimeout)
+		defer cancel()
+		req, err := qr.stmt.conn.newRequest(ctx, "DELETE", qr.nextURI, nil, hs)
+		if err != nil {
+			return err
 		}
-		return err
+		resp, err := qr.stmt.conn.roundTrip(ctx, req)
+		if err != nil {
+			qferr, ok := err.(*ErrQueryFailed)
+			if ok && qferr.StatusCode == http.StatusNoContent {
+				qr.nextURI = ""
+				return nil
+			}
+			return err
+		}
+		resp.Body.Close()
+
 	}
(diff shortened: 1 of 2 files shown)


nineinchnick:
@cla-bot check

Flgado:
@nineinchnick  add a need unit test. I think it make sense

nineinchnick:
@cla-bot check

trino/trino.go
```diff
@@ -1138,27 +1138,34 @@ func (qr *driverRows) Close() error {
...
-		return err
-	}
-	resp, err := qr.stmt.conn.roundTrip(ctx, req)
-	if err != nil {
-		qferr, ok := err.(*ErrQueryFailed)
-		if ok && qferr.StatusCode == http.StatusNoContent {
-			qr.nextURI = ""
-			return nil
+
+	if qr.nextURI != "" {
```
nineinchnick, line 1144: Reverse this condition and do an early return to avoid indenting the remaining code.

trino/trino.go
```diff
@@ -1215,22 +1223,34 @@ func (qr *driverRows) Next(dest []driver.Value) error {
...
 	}
+
 	if len(qr.coltype) == 0 {
 		qr.err = sql.ErrNoRows
 		return qr.err
 	}
-	for i, v := range qr.coltype {
-		if i > len(dest)-1 {
+
+	row := qr.data[qr.rowindex]
```
nineinchnick, line 1232: Move any refactor not directly related to the issue this PR fixes to a separate commit.

trino/trino.go
```diff
@@ -1215,22 +1223,34 @@ func (qr *driverRows) Next(dest []driver.Value) error {
...
 			qr.err = err
 			return err
 		}
-		dest[i] = vv
+		dest[i] = val
 	}
+
 	qr.rowindex++
+
+	// Prefetch next set of rows
```
nineinchnick, line 1247: Useless comment - it explains the what, which is pretty obvious, but not the why, which is not. Why do we need to do this? If the user doesn't request any more data, we'll waste resources on one round-trip.

Can we do this only when cancelling the query?

trino/trino_test.go
```diff
@@ -1973,3 +1973,198 @@ func TestForwardAuthorizationHeader(t *testing.T) {
 
 	assert.NoError(t, db.Close())
 }
+
+func TestPagination(t *testing.T) {
```
nineinchnick, line 1977: Can you come up with a better name? I'm not sure what this test is for.

trino/trino_test.go
```diff
@@ -1973,3 +1973,198 @@ func TestForwardAuthorizationHeader(t *testing.T) {
...
+					},
+				})
+			}
+			w.WriteHeader(http.StatusOK)
+			w.Write(buf.Bytes())
+			return
+		}
+
+		if r.URL.Path == "/v1/statement/20210817_140827_00000_arvdv/1" {
+			if buf2 == nil {
```
nineinchnick, line 1999: Why are separate buffers needed here? Can we initialize them early, and remove this condition?

Issue #131, opened 2024-12-17
Context Deadline should be configurable


lorransr:
I created this [issue](https://github.com/xo/usql/issues/506#issuecomment-2548873490) on usql. I found that for long queries the context deadline is kind of short and I believe it should be configurable. Is it possible? and does it makes sense?

KasonBraley:
I have encountered this recently as well, and was quite surprised by this default. Would you be open to removing this default query timeout entirely and pushing that responsibility onto the user of the package? If a user would want to set a timeout or deadline on the query, they could just set that accordingly on the context that they pass in.

nineinchnick:
We should already handle a timeout set in a context. The default timeout can be increased by changing the [trino.DefaultQueryTimeout](https://github.com/trinodb/trino-go-client/blob/master/trino/trino.go#L87) variable, but it's global, cannot be changed through the URL, and generally it's a poor design choice. I'd suggest increasing it significantly, and deprecate it first, just in case anyone is relying on it.

Chaho12:
@lorransr fyi. you can set query timeout by passing context like this.

```go
	timeoutCtx, cancelQueryContext := context.WithTimeout(*c.CancelCtx, 120*time.Second)
	defer cancelQueryContext()

	rows, err := qh.Connection.QueryContext(timeoutCtx, query,
```

PR #137, opened 2025-03-13
Add support for optional timeout param in the DSN

aalan3:
This PR adds `timeout` to the DSN. Currently it's only possible to extend the timeout via the context, otherwise it's 1 min via `DefaultQueryTimeout`. 

Related to https://github.com/trinodb/trino-go-client/issues/131

### DIFF
--- a/README.md
+++ b/README.md
@@ -239,6 +239,16 @@ trino.RegisterCustomClient("otel", otelClient)
 db, err := sql.Open("trino", "https://user@localhost:8080?custom_client=otel")
 ```
 
+##### `queryTimeout`
+
+```
+Type:           time.Duration
+Valid values:   duration string
+Default:        nil
+```
+
+The `queryTimeout` parameter sets a timeout for the query. If the query takes longer than the timeout, it will be cancelled. If it is not set the default context timeout will be used.
+
 #### Examples
 
 ```
--- a/trino/trino.go
+++ b/trino/trino.go
@@ -85,7 +85,7 @@ func init() {
 
 var (
 	// DefaultQueryTimeout is the default timeout for queries executed without a context.
-	DefaultQueryTimeout = 60 * time.Second
+	DefaultQueryTimeout = 10 * time.Hour
 
 	// DefaultCancelQueryTimeout is the timeout for the request to cancel queries in Trino.
 	DefaultCancelQueryTimeout = 30 * time.Second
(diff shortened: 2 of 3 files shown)


trino/trino.go
```diff
@@ -963,9 +979,12 @@ func (st *driverStmt) exec(ctx context.Context, args []driver.NamedValue) (*stmt
 	}
 
 	var cancel context.CancelFunc = func() {}
-	if _, ok := ctx.Deadline(); !ok {
+	if st.conn.timeout != nil {
+		ctx, cancel = context.WithTimeout(ctx, *st.conn.timeout)
+	} else if _, ok := ctx.Deadline(); !ok {
```
nineinchnick, line 984: Both timeouts should always apply

README.md
```diff
@@ -239,6 +239,16 @@ trino.RegisterCustomClient("otel", otelClient)
...
 
+##### `timeout`
+
+```
+Type:           time.Duration
+Valid values:   duration string
+Default:        nil
+```
+
+The `timeout` parameter sets a timeout for the query. If the query takes longer than the timeout, it will be cancelled. If it is not set the default context timeout will be used.
```
nineinchnick, line 250: Call this queryTimeout

trino/trino.go
aalan3: Not sure I understand, if both timeouts apply wouldn't this be capped at 60s? Should I increase the default in that case?

nineinchnick: If a user sets timeouts both in the connection string, and in the context, we should never silently ignore one of them. We should choose the smaller one. The default only applies when no timeout is set.

I do think we should increase the default significantly, maybe even to 10 hours. Small timeouts are fine for things like network requests, which can get stuck, and can be retried. We're in no position to assume there's anything wrong with a query that runs for a long time.

nineinchnick: Ah that's already how your changes work - I think adding a timeout to a context with an existing deadline will cause one to trigger first anyway.

nineinchnick:
Please squash commits, this is ready to merge once the CLA clears out. I'll keep an eye on that. Thanks!

nineinchnick (approved):

nineinchnick:
@cla-bot check

aalan3:
@cla-bot check

PR #138, opened 2025-03-14
Spooling Protocol

Flgado:
## Description
his PR introduces the initial implementation of the `spooling protocol` in the Trino Go client. Currently, the Go client lacks support for spooling, which I believe is a highly valuable feature for the entire community.

For this first iteration, the implementation follows a straightforward approach, ensuring compatibility with the existing direct protocol. In the future, we could explore optimizations such as fetching segments in parallel and processing them out of order when necessary.

This PR is `not` a final version but rather a starting point for discussion around the implementation, testing strategy, and potential improvements.

## Implementation Details
- Implemented the spooling protocol while ensuring full `backward compatibility` with the direct protocol.
- Based the implementation on the `Python client` and relevant `Trino documentation`.
- Added initial `integration tests` to cover various spooling protocol use cases.
- While the current tests validate core functionality, I believe we need additional coverage to ensure correctness across all edge cases (see concerns below).

### Performance
- `Efficiency is critical` in Trino use cases. This implementation makes use of compression algorithms and frequently allocates memory slices.
- We should analyze whether we can `reduce unnecessary allocations` and optimize our usage of the `LZ4` and `Zstd` decompression algorithms for better performance.
### Testing
- The integration tests currently cover both `inline` segments` and `spooled segments`, but the behavior depends on Trino configurations and the size of returned values.
  - How can we ensure we are consistently testing the right use cases across different configurations?
- It would be `highly beneficial` if an experienced community member familiar with the spooling protocol could help identify missing test cases and contribute additional unit tests for broader coverage.


trino/integration_test.go
```diff
@@ -1325,3 +1460,240 @@ func TestIntegrationLargeQuery(t *testing.T) {
...
+	}
+	if rows.Err() != nil {
+		t.Fatal(err)
+	}
+	if count != 100 {
+		t.Fatal("not enough rows returned:", count)
+	}
+}
+
+func TestIntegrationSelectTpch100SpoolingInlineSegmentJsonlz4Encoded(t *testing.T) {
```
nineinchnick, line 1667: The names of these tests should help developers understand why they're needed, they don't have to list all options used. If you're testing all supported compression methods, I'd suggest doing this in a single, tabular test, with a name that'd say so.

trino/integration_test.go
```diff
@@ -1325,3 +1460,240 @@ func TestIntegrationLargeQuery(t *testing.T) {
...
+			ARRAY[1, 2, NULL],
+			ARRAY[ARRAY[1, 1, 1], NULL],
+			ARRAY[ARRAY[ARRAY[1, 1, 1], NULL], NULL],
+			ARRAY[1.0, 2.0, NULL],
+			ARRAY[ARRAY[1.1, 1.1, 1.1], NULL],
+			ARRAY[ARRAY[ARRAY[1.1, 1.1, 1.1], NULL], NULL],
+			MAP(ARRAY['a', 'b'], ARRAY['c', 'd']),
+			CAST(NULL AS MAP(ARRAY(INTEGER), ARRAY(INTEGER))),
+			ROW(1, 'a', CAST('2017-07-10 01:02:03.004 UTC' AS TIMESTAMP(6) WITH TIME ZONE), ARRAY['c'])
+	`, sql.Named("X-Trino-Query-Data-Encoding", "json")).Scan(
```
nineinchnick, line 1508: Are you sure this is testing spooling? If yes, can you document why?

trino/trino.go
```diff
@@ -1253,13 +1263,192 @@ type queryResponse struct {
...
+}
+
+func (sp *spoolingProtocol) fetchSegment(uri, ackUri string, headers map[string]interface{}, metaData map[string]interface{}) ([]queryData, error) {
+	req, err := http.NewRequestWithContext(sp.ctx, "GET", uri, nil)
+	if err != nil {
+		return nil, fmt.Errorf("trino: %w", err)
+	}
+
+	for k, v := range headers {
+		val := reflect.TypeOf(v)
```
nineinchnick, line 1325: Can we avoid using reflection, and only use type assertion?

trino/trino.go
```diff
@@ -1253,13 +1263,192 @@ type queryResponse struct {
...
+		}
+
+		resp, err := sp.httClient.Do(ackReq)
+		if err != nil {
+			return
+		}
+		resp.Body.Close()
+	}()
+
+	return decode(data, sp.encoder, metaData)
```
nineinchnick, line 1378: Ideally, fetching and decoding should be done in parallel, same as we're already doing this for the direct protocol. Can you prepare a benchmark to compare the two implementations? We have one for the direct protocol: https://github.com/trinodb/trino-go-client/blob/master/trino/trino_test.go#L1855

Flgado: make sense 👍🏻

trino/trino.go
```diff
@@ -1253,13 +1263,193 @@ type queryResponse struct {
...
-	Data             []queryData   `json:"data"`
+	Data             interface{}   `json:"data"`
 	Stats            stmtStats     `json:"stats"`
 	Error            ErrTrino      `json:"error"`
 	UpdateType       string        `json:"updateType"`
 	UpdateCount      int64         `json:"updateCount"`
 }
 
+type spoolingProtocol struct {
+	httClient http.Client
```
wendigo, line 1274: nit httpClient

trino/trino.go
```diff
@@ -1253,13 +1263,193 @@ type queryResponse struct {
...
+	ctx       context.Context
+	encoder   string
```
wendigo, line 1276: encoding

trino/trino.go
```diff
@@ -1253,13 +1263,193 @@ type queryResponse struct {
...
+	segments  []interface{}
+}
+
+func (sp *spoolingProtocol) fetch() ([]queryData, error) {
+	var queryData []queryData
+	for _, segment := range sp.segments {
+		segment := segment.(map[string]interface{})
+		metaData, _ := segment["metadata"].(map[string]interface{})
```
wendigo, line 1284: nit: `metadata, _ :=`

trino/trino.go
```diff
@@ -1253,13 +1263,193 @@ type queryResponse struct {
...
+			uri, _ := segment["uri"].(string)
+			ackUri, _ := segment["ackUri"].(string)
+			headers, _ := segment["headers"].(map[string]interface{})
+
+			segmentdata, err := sp.fetchSegment(uri, ackUri, headers, metaData)
+			if err != nil {
+				return nil, err
+			}
+
+			queryData = append(queryData, segmentdata...)
```
wendigo, line 1310: can we make it lazy?

trino/trino.go
```diff
@@ -1253,13 +1263,193 @@ type queryResponse struct {
...
+		}
+
+		resp, err := sp.httClient.Do(ackReq)
+		if err != nil {
+			return
+		}
+		resp.Body.Close()
+	}()
+
+	return decode(data, sp.encoder, metaData)
```
wendigo, line 1373: for consistency I'd return bytes here and move decoding to the `fetch` method

wendigo (commented):
Skimmed

trino/trino.go
wendigo: Works for me!

Flgado:
@cla-bot check

trino/etc/config-pre-466version.properties
```diff
@@ -0,0 +1,15 @@
...
+
+http-server.authentication.type=PASSWORD,JWT
+http-server.authentication.jwt.key-file=/etc/trino/secrets/public_key.pem
+http-server.https.enabled=true
+http-server.https.port=8443
+http-server.authentication.allow-insecure-over-http=true
+http-server.https.keystore.path=/etc/trino/secrets/certificate_with_key.pem
+internal-communication.shared-secret=gotrino
+
+query.max-length=5000043
```
wendigo, line 15: add trailing new line

trino/etc/config.properties
```diff
@@ -14,3 +14,8 @@ http-server.https.keystore.path=/etc/trino/secrets/certificate_with_key.pem
 internal-communication.shared-secret=gotrino
 
 query.max-length=5000043
+
+## spooling protocol settings
+protocol.spooling.enabled=true
+protocol.spooling.shared-secret-key=jxTKysfCBuMZtFqUf8UJDQ1w9ez8rynEJsJqgJf66u0=
+protocol.spooling.retrieval-mode=coordinator_proxy
```
wendigo, line 21: add trailing new line

trino/trino.go
```diff
@@ -1253,13 +1263,233 @@ type queryResponse struct {
...
+	for _, segment := range sp.segments {
+		segment := segment.(map[string]interface{})
+		metadataAssertion, _ := segment["metadata"].(map[string]interface{})
+		metadata, err := parseSpoolingMetadata(metadataAssertion)
+		if err != nil {
+			return nil, err
+		}
+		switch segment["type"] {
+		case "inline":
+			data := segment["data"].(string)
```
wendigo, line 1342: you can inline data in the call next line

trino/etc/config.properties
Flgado: Only to simplify the integration test configurations, Trino and LocalStack are running inside Docker on the same network. This allows Trino to easily find and communicate with LocalStack by calling http://localstack:4566/. However, if I use storage mode (directly calling storage), Trino will respond with http://localstack:port/spooling/..., and the integration tests won't be able to call it, because the tests are running locally.

trino/integration_test.go
```diff
@@ -155,16 +218,122 @@ func TestMain(m *testing.M) {
...
+	}
+
+	// Create S3 bucket for integration tests
+	for i := 0; i < 10; i++ {
+		err := runLocalStackCommand("test", "test", fmt.Sprintf("aws --endpoint-url=%s s3 mb s3://%s", s3Endpoint, bucketName))
+		if err == nil {
+			log.Println("S3 bucket created successfully")
+			return s3Endpoint, nil
+		}
+		log.Printf("Failed to create S3 bucket, retrying... (%d/10)\n", i+1)
```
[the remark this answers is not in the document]
Flgado: Yes, I haven’t done it to avoid adding an extra dependency on the client. I can change it if you prefer using the s3 client. Let me know 👍🏻

trino/trino.go
```diff
@@ -1253,13 +1264,239 @@ type queryResponse struct {
...
+		}
+
+		for k, values := range req.Header {
+			for _, v := range values {
+				ackReq.Header.Add(k, v)
+			}
+		}
+
+		resp, err := sp.httpClient.Do(ackReq)
+		if err != nil {
```
nineinchnick, line 1434: We should handle these errors somehow. Even if they're not actionable (acks are optional), there should be a way for the users to be aware that some acks are failing. We could send them to some buffered channel that would collect them and either return to the user or discard when the whole statement is being closed. Only ignore them when the buffered channel is full.

This could be done in the same way how the driver allows monitoring the query progress via a callback.

Since it sounds like a significant amount of work, we can do it as a follow up.

nineinchnick:
@Flgado did you sent in the signed CLA? The last batch was processed yesterday, and I don't see you on the list. I won't be able to merge it without it.

trino/integration_test.go
```diff
@@ -1325,3 +1509,150 @@ func TestIntegrationLargeQuery(t *testing.T) {
 		t.Fatal("not enough rows returned:", count)
 	}
 }
+
+func TestIntegrationTypeConversionSpoolingProtocolInlineJsonEncoder(t *testing.T) {
```
nineinchnick, line 1513: How is this testing the spooling protocol? I don't see any extra headers

trino/integration_test.go
```diff
@@ -1325,3 +1509,150 @@ func TestIntegrationLargeQuery(t *testing.T) {
...
+			expected: 1000,
+		},
+		{
+			name:     "Spooling Spooled Segment JSON+LZ4 Encoded",
+			query:    "SELECT * FROM tpch.sf1.customer LIMIT 1000",
+			encoding: "json+lz4",
+			expected: 1000,
+		},
+		{
+			name:     "Spooling Inline Segment JSON+ZSTD Encoded",
```
nineinchnick, line 1606: How do you know segments are inlined?

trino/trino.go
```diff
@@ -1253,13 +1264,263 @@ type queryResponse struct {
...
+	RowsCount        int64
+	SegmentSize      int64
+	UncompressedSize int64
+}
+
+func parseSpoolingMetadata(metadata map[string]interface{}) (spoolingMetadata, error) {
+	var mt spoolingMetadata
+
+	// mandatory field
+	if val, ok := metadata["rowOffset"].(json.Number); ok {
```
nineinchnick, line 1292: `ok` is only applied on the type assertion. Can you add a separate check to make sure the `rowOffse` key is present?

trino/trino.go
```diff
@@ -1253,13 +1264,263 @@ type queryResponse struct {
...
+		num, ok := val.(json.Number)
+		if !ok {
+			return spoolingMetadata{}, fmt.Errorf("invalid type for uncompressedSize on segment metadata, expected json.Number")
+		}
+		var err error
+		if mt.UncompressedSize, err = num.Int64(); err != nil {
+			return spoolingMetadata{}, fmt.Errorf("error converting uncompressedSize to int64: %v", err)
+		}
+	} else {
+		mt.UncompressedSize = 0
```
nineinchnick, line 1323: Set the default at the top of this method, when you initialize `mt`.

trino/trino.go
```diff
@@ -1253,13 +1264,263 @@ type queryResponse struct {
...
+
+type spoolingMetadata struct {
+	RowOffset        int64
+	RowsCount        int64
+	SegmentSize      int64
+	UncompressedSize int64
+}
+
+func parseSpoolingMetadata(metadata map[string]interface{}) (spoolingMetadata, error) {
+	var mt spoolingMetadata
```
nineinchnick, line 1289: Use more meaningful variable names, not just abbreviations.
```suggestion
	var result spoolingMetadata
```

trino/trino.go
```diff
@@ -1253,13 +1264,263 @@ type queryResponse struct {
...
+func (sp *spoolingProtocol) fetch() ([]queryData, error) {
+	var queryData []queryData
+	for segmentIndex, segment := range sp.segments {
+		segment, ok := segment.(map[string]interface{})
+		if !ok {
+			return nil, fmt.Errorf("segment at index %d is invalid: expected map[string]interface{}, got %T", segmentIndex, segment)
+		}
+		typedMetadata, ok := segment["metadata"].(map[string]interface{})
+		if !ok {
+			return nil, fmt.Errorf("metadata missing or invalid in segment at index %d", segmentIndex)
```
nineinchnick, line 1352: Same here, we only know the type was invalid, not if the field was missing.

trino/trino.go
```diff
@@ -1253,13 +1264,263 @@ type queryResponse struct {
...
+			}
+
+			queryData = append(queryData, decodedData...)
+		case "spooled":
+			uri, ok := segment["uri"].(string)
+			if !ok || uri == "" {
+				return nil, fmt.Errorf("missing or invalid 'uri' field in spooled segment at index %d", segmentIndex)
+			}
+			ackUri, ok := segment["ackUri"].(string)
+			if !ok || ackUri == "" { // Is mandatory?
```
nineinchnick, line 1379: Remove these comments, let's assume these fields are required

trino/trino.go
```diff
@@ -1253,13 +1264,263 @@ type queryResponse struct {
...
+		headerSlice, ok := v.([]interface{})
+		if !ok {
+			return nil, fmt.Errorf("unsupported header type %T", v)
+		}
+
+		if len(headerSlice) > 1 {
+			return nil, fmt.Errorf("multiple values for header %s", k)
+		}
+
+		header, ok := headerSlice[0].(string)
```
nineinchnick, line 1421: Ignore empty `headerSlice`

trino/trino.go
```diff
@@ -1253,13 +1264,263 @@ type queryResponse struct {
...
+		if !ok {
+			return nil, fmt.Errorf("unsupported header value type %T for spooled segment", headerSlice[0])
```
nineinchnick, line 1423: You wrap this error later, so drop `for spooled segment`.

trino/trino.go
```diff
@@ -1253,13 +1264,263 @@ type queryResponse struct {
...
+		resp.Body.Close()
+	}()
+
+	return data, nil
+}
+
+func decode(data []byte, encoding string, metadata spoolingMetadata) ([]queryData, error) {
+	if metadata.UncompressedSize == 0 {
+		var queryDataList = make([]queryData, metadata.RowsCount)
+		decoder := json.NewDecoder(bytes.NewReader(data))
```
nineinchnick, line 1470: We should not duplicate this block of code (json decoding). Extract decompression to a method, and call it first in this method.

trino/trino_test.go
```diff
@@ -1147,6 +1154,656 @@ func TestFetchNoStackOverflow(t *testing.T) {
...
+			for rows.Next() {
+				var value int
+				err := rows.Scan(&value)
+				require.NoError(t, err)
+				results = append(results, value)
+			}
+
+			require.NoError(t, rows.Err())
+
+			expected := []int{1000, 10001}
```
nineinchnick, line 1277: Use expected values from `tc`.

trino/trino_test.go
```diff
@@ -1147,6 +1154,656 @@ func TestFetchNoStackOverflow(t *testing.T) {
...
+		Segments       []map[string]interface{}
+		ExpectedResult []int
+		Encoding       string
+	}{
+		{
+			Name: "noCompression",
+			Segments: []map[string]interface{}{
+				{
+					"type":     "inline",
+					"data":     jsonWithoutCompression,
```
nineinchnick, line 1304: Inline all of these

nineinchnick (approved):
Looks great, I really appreciate the tests for error handling. I added a few nitpick comments, but this is ready to be merged.

trino/integration_test.go
Flgado: I selected fewer rows to stay below the `protocol.spooling.inlining.max-rows` and `protocol.spooling.inlining.max-size` defaults (1000 rows and 128kB), which causes the segments to be inlined instead of spooled. And of course I validate that the response where inline segments and not spooled for this cases.

Let me know if you’d like me to add a comment in the test to make that behavior explicit.

trino/trino.go
Flgado: thanks 👍🏻

trino/integration_test.go
Flgado: Is before the scan. `sql.Named(trinoEncoding, "json"))`

nineinchnick:
@cla-bot check

nineinchnick:
Squash all commits and I'll merge this, thank you!

Flgado:
@nineinchnick  done

wendigo:
I've amended commit message and going to merge once the CI is done

wendigo (approved):

nineinchnick (approved):

wendigo:
Great work @Flgado! Thanks for this contribution :)

PR #139, opened 2025-03-25
Parallel Processing of Spooled Segments

Flgado:
## Description
This PR continues the work from [#138](https://github.com/trinodb/trino-go-client/pull/138).

The goal of this PR is to leverage the spooling protocol's behavior to improve performance. It introduces parallelism in the processing of spooled segments, resulting in a noticeable performance improvement in benchmark tests.

I chose a fan-out approach with two workers because, after running benchmarks and analyzing different strategies, this proved to be the most efficient. Additionally, this solution allows for future configurability, enabling the number of workers to be adjusted per request if needed and if we decide to go in that path.

## What is missing
- [ ] I initially assumed that `rowsCount` was a mandatory field, but I discovered that this is not the case for inline segments. I have reached out to the core developers in the community channel to clarify this behavior.


trino/trino.go
```diff
@@ -1253,13 +1263,403 @@ type queryResponse struct {
...
+		combinedQueryData = append(combinedQueryData, segmentData...)
+	}
+
+	return combinedQueryData, nil
+}
+
+// processSegments processes data segments and sends HTTP responses through a channel.
+//
+// It iterates over `sp.segments`, extracting metadata and handling each segment based on its type:
+// - "inline" segments are decoded and stored directly in `processedSegments`.
```
nineinchnick, line 1503: If the words don't mean something else, don't double quote them.

trino/trino.go
```diff
@@ -1253,13 +1263,403 @@ type queryResponse struct {
...
+				}
+
+				decodedData, err := decode(decodedBytes, sp.encoder, metaData)
+				if err != nil {
+					ch <- spooledSegToProccess{
+						err: err,
+					}
+				}
+
+				processedSegments[i] = decodedData
```
nineinchnick, line 1550: Why mutate the method parameter, instead of sending something to the channel? If this is correct, explain it in a code comment.

trino/trino.go
```diff
@@ -1253,13 +1263,403 @@ type queryResponse struct {
...
+func (sp *spoolingProtocol) fetchSegment(uri, ackUri string, headers map[string]interface{}, metaData spoolingMetaData) spooledSegToProccess {
+	req, err := http.NewRequestWithContext(sp.ctx, "GET", uri, nil)
+	if err != nil {
+		return spooledSegToProccess{err: fmt.Errorf("trino: %w", err)}
+	}
+
+	for k, v := range headers {
+		headerSlice, ok := v.([]interface{})
+
+		if ok {
```
nineinchnick, line 1574: Reverse the condition and do an early return.

trino/trino.go
```diff
@@ -1253,13 +1263,403 @@ type queryResponse struct {
...
+		}
+	}()
+
+	return ch
+}
+
+func (sp *spoolingProtocol) fetchSegment(uri, ackUri string, headers map[string]interface{}, metaData spoolingMetaData) spooledSegToProccess {
+	req, err := http.NewRequestWithContext(sp.ctx, "GET", uri, nil)
+	if err != nil {
+		return spooledSegToProccess{err: fmt.Errorf("trino: %w", err)}
```
nineinchnick, line 1568: Make the error message actionable. Why there's a `trino:` prefix in all errors in this method?

trino/trino.go
```diff
@@ -1253,13 +1263,403 @@ type queryResponse struct {
...
+	ackHeaders := make(http.Header)
+	for k, values := range req.Header {
+		for _, v := range values {
+			ackHeaders.Add(k, v)
+		}
+	}
+
+	return spooledSegToProccess{
+		Response:   resp,
+		MetaData:   metaData,
```
nineinchnick, line 1599: Why this method just copy the metadata and the ack uri and headers? Maybe it should only return the response?

trino/trino.go
```diff
@@ -1253,13 +1263,403 @@ type queryResponse struct {
...
+
+func decode(data []byte, encoder string, metaData spoolingMetaData) ([]queryData, error) {
+
+	if metaData.UncompressedSize == 0 {
+		var queryDataList []queryData
+		decoder := json.NewDecoder(bytes.NewReader(data))
+		decoder.UseNumber()
+		err := decoder.Decode(&queryDataList)
+		if err != nil {
+			return nil, fmt.Errorf("error unmarshalling JSON: %v", err)
```
nineinchnick, line 1613: Make the error more actionable - what JSON?

trino/trino.go
```diff
@@ -1253,13 +1263,403 @@ type queryResponse struct {
...
+		return nil, fmt.Errorf("segment size mismatch: expected %d bytes, got %d bytes", metaData.SegmentSize, len(data))
+	}
+
+	var decompressedData []byte
+	var err error
+	switch encoder {
+	case "json+zstd":
+		zstdReader, err := zstd.NewReader(bytes.NewReader(data))
+		if err != nil {
+			return nil, fmt.Errorf("trino: error creating zstd reader: %w", err)
```
nineinchnick, line 1628: Why the `trino:` prefix?

trino/trino.go
```diff
@@ -1326,12 +1726,39 @@ func (qr *driverRows) fetch() error {
...
 			}
+
 			qr.rowindex = 0
-			qr.data = qresp.Data
+			switch data := qresp.Data.(type) {
+			case []interface{}:
+				// direct protocol
+				qr.data = make([]queryData, len(data))
+				for i, item := range data {
+					if row, ok := item.([]interface{}); ok {
```
nineinchnick, line 1741: Should we throw an error if the type assertion fails?

trino/integration_test.go
```diff
@@ -1325,3 +1460,240 @@ func TestIntegrationLargeQuery(t *testing.T) {
...
+	}
+	if count != 100 {
+		t.Fatal("not enough rows returned:", count)
+	}
+}
+
+func TestIntegrationSelectTpch100SpoolingInlineSegmentJsonlz4Encoded(t *testing.T) {
+	db := integrationOpen(t)
+	defer db.Close()
+	rows, err := db.Query("SELECT * FROM tpch.sf1.customer LIMIT 100", sql.Named("X-Trino-Query-Data-Encoding", "json+lz4"))
```
nineinchnick, line 1670: We need a test that will make sure results are returned in order. Generate a sequence of numbers and assert that it's monotonic (every value is greater than the previous one).

trino/trino.go
```diff
@@ -1253,13 +1263,403 @@ type queryResponse struct {
...
+
+	errorChannel := make(chan error)
+	defer close(errorChannel)
+
+	fetchContext, cancelFetch := context.WithCancel(context.Background())
+	defer cancelFetch()
+
+	spoolingSegmentProcessors := 1
+	if numberOfSeg >= 2 {
+		spoolingSegmentProcessors = 2
```
nineinchnick, line 1456: How did you choose this number?

trino/trino.go
```diff
@@ -1253,13 +1263,403 @@ type queryResponse struct {
...
+
+	spooledSegmentStream := sp.processSegments(processedDataSegments, rowOffsetToSegmentIndex, totalRowsCount)
+
+	for range spoolingSegmentProcessors {
+		workersCloseSignals = append(workersCloseSignals, sp.proccessSpolledSegment(fetchContext, spooledSegmentStream, errorChannel, processedDataSegments, rowOffsetToSegmentIndex))
+	}
+
+	var fetchError error
+
+	finishedWorkersIndex := 0
```
nineinchnick, line 1469: Use https://pkg.go.dev/sync#WaitGroup instead

nineinchnick (commented):
This adds some parallelism, but building the list of segments (iterating over query results by following the nextUri) should be completely separate from downloading of segments. Currently, the driver will wait before getting the next list of segments until all current segments are downloaded.

It'll be more complicated to start multiple goroutines to download segments and preserve the order without allocating memory for the complete result set.

There should be a buffered channel to collect all segments metadata. If all segment metadata fits, the query will be finished on the server, releasing resources. 

Assuming two buffered input and output channels with size of N, we should:
1. Create N worker goroutines
2. Each worker reads segment metadata from the input, downloads and decodes it, and writes to the output, together with the worker number, blocking if there's nothing to read.
3. The main goroutine writes N segment metadata to the input channel, reads N segments from the output, sorts it by worker numbers, and makes it available as `qr.data` for the driver client to process.
4. Once the driver client reads all `qr.data`, download the next batch of segments - repeat from step 3.

Once this is implemented, we should also add keepalives. If all segments metadata doesn't fit in the first channel, the goroutine that tries to write to it should also watch a timer and send a periodical HEAD request to the last nextUri, to keep the query alive on the server.

trino/trino.go
Flgado: saw that we use the convention in some cases. Removed

trino/trino.go
```diff
@@ -1253,13 +1263,403 @@ type queryResponse struct {
...
+	}
+
+	segmentSize, err := segmentSizeAssertion.Int64()
+	if err != nil {
+		return spoolingMetaData{}, fmt.Errorf("error converting segmentSize to int64")
+	}
+
+	var uncompressedSize int64
+	if uncompressedSizeAssertion, ok := metaData["uncompressedSize"].(json.Number); ok {
+		if val, err := uncompressedSizeAssertion.Int64(); err == nil {
```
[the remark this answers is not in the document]
Flgado: I am only not ignoring the ones that are mandatory fields on metada. In this case I ignoring this one because it could not exist, it only exist when the segment is compressed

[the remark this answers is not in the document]
nineinchnick: You can ignore a missing value, but not an invalid value. This would indicate some serious issues with the server, and the clients using this driver should be aware of that.

PR #140, opened 2025-04-01
Ensure Query Cancellation on Context Deadline Expiry 

Flgado:
## Problem description
- When the driver is used with a context deadline, and the deadline is reached before processing all data, the driver does not cancel the underlying Trino query. This happens because the expiration of the context deadline isn't properly handled, and the driver does not trigger the necessary steps to cancel the query.

## Solution
This pull request addresses the issue by ensuring that the Close method is called when the context deadline is reached. This method will initiate a delete request, notifying Trino to cancel the corresponding underlying query. By adding this behavior, the query will be properly canceled when the context deadline expires, preventing unnecessary resource usage and improving the overall reliability of the system.

### DIFF
--- a/trino/integration_test.go
+++ b/trino/integration_test.go
@@ -1008,82 +1008,118 @@ func TestIntegrationUnsupportedHeader(t *testing.T) {
 	}
 }
 
-func TestIntegrationQueryContextCancellation(t *testing.T) {
-	err := RegisterCustomClient("uncompressed", &http.Client{Transport: &http.Transport{DisableCompression: true}})
-	if err != nil {
+func TestIntegrationQueryContext(t *testing.T) {
+	tests := []struct {
+		name           string
+		timeout        time.Duration
+		expectedErrMsg string
+	}{
+		{
+			name:           "Context Cancellation",
+			timeout:        0,
+			expectedErrMsg: "canceled",
+		},
+		{
+			name:           "Context Deadline Exceeded",
+			timeout:        3 * time.Second,
+			expectedErrMsg: "context deadline exceeded",
+		},
+	}
+
+	if err := RegisterCustomClient("uncompressed", &http.Client{Transport: &http.Transport{DisableCompression: true}}); err != nil {
 		t.Fatal(err)
 	}
-	dsn := *integrationServerFlag
-	dsn += "?catalog=tpch&schema=sf100&source=cancel-test&custom_client=uncompressed"
+
+	dsn := *integrationServerFlag + "?catalog=tpch&schema=sf100&source=cancel-test&custom_client=uncompressed"
 	db := integrationOpen(t, dsn)
 	defer db.Close()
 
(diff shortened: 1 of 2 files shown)


trino/trino.go
```diff
@@ -1344,6 +1344,13 @@ func (qr *driverRows) fetch() error {
...
 			} else if err == context.Canceled {
 				qr.Close()
+			} else {
+				var queryErr *ErrQueryFailed
+				if errors.As(err, &queryErr) {
+					if errors.Is(queryErr.Reason, context.DeadlineExceeded) {
+						qr.Close()
+					}
+				}
 			}
```
nineinchnick, line 1354: This can be simplified because `ErrQueryFailed` implements the unwrap interface. From https://pkg.go.dev/errors:
> [Is](https://pkg.go.dev/errors#Is) examines the tree of its first argument looking for an error that matches the second. It reports whether it finds a match. It should be used in preference to simple equality checks

```suggestion
			} else if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
				qr.Close()
			}
```

trino/integration_test.go
```diff
@@ -1008,82 +1008,123 @@ func TestIntegrationUnsupportedHeader(t *testing.T) {
...
-		if err != sql.ErrNoRows {
-			t.Fatal("failed to read query id", err)
-		}
-		if err = contextSleep(pollCtx, 100*time.Millisecond); err != nil {
-			t.Fatal("query did not start in 1 second")
-		}
-	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := RegisterCustomClient("uncompressed", &http.Client{Transport: &http.Transport{DisableCompression: true}})
```
nineinchnick, line 1034: Pull all the code common to all tests before the loop. It will make the diff smaller, and make the test easier to read - the loop will only contain the test action and assertion, not the setup.

trino/integration_test.go
```diff
@@ -1008,82 +1008,123 @@ func TestIntegrationUnsupportedHeader(t *testing.T) {
...
-		if err != nil {
-			t.Fatal("failed to read query id", err)
-		}
-		if state == "FAILED" && code != nil && *code == "USER_CANCELED" {
-			break
-		}
-		if err = contextSleep(pollCtx, 100*time.Millisecond); err != nil {
-			t.Fatal("query was not cancelled in 1 second; state, code, err are:", state, code, err)
-		}
+			if tt.manualCancel {
```
nineinchnick, line 1047: I don't think we need a separate field like this, it would be enough to check if timeout is set (is non-zero).

nineinchnick (approved):
Just some nitpicks, looks good otherwise. Thanks for working on this!

nineinchnick:
@cla-bot check

nineinchnick:
@Flgado can you squash both commits? This is ready to be merged

nineinchnick (approved):

PR #141, opened 2025-04-29
Spooling parallelization

Flgado:
# Description
This PR introduces parallel downloading of all segments on the client side, while ensuring that the main goroutines responsible for fetching additional metadata from Trino remain non-blocking.
By enabling concurrent segment downloads, we achieve improved performance and faster metadata retrieval, resulting in more efficient resource usage and quicker release of Trino coordinator resources.

### DIFF
--- a/trino/trino.go
+++ b/trino/trino.go
@@ -107,6 +107,10 @@ var (
 
 	// ErrInvalidProgressCallbackHeader indicates that server did not get valid headers for progress callback
 	ErrInvalidProgressCallbackHeader = errors.New("trino: both " + trinoProgressCallbackParam + " and " + trinoProgressCallbackPeriodParam + " must be set when using progress callback")
+
+	DefaultSpoolingDownloadWorkers = 5
+
+	DefaultSpoolingDecodersWorkers = 5
 )
 
 const (
@@ -690,15 +694,127 @@ func newErrQueryFailedFromResponse(resp *http.Response) *ErrQueryFailed {
 }
 
 type driverStmt struct {
-	conn           *Conn
-	query          string
-	user           string
-	nextURIs       chan string
-	httpResponses  chan *http.Response
-	queryResponses chan queryResponse
-	statsCh        chan QueryProgressInfo
-	errors         chan error
-	doneCh         chan struct{}
+	conn                        *Conn
+	query                       string
+	user                        string
+	nextURIs                    chan string
+	httpResponses               chan *http.Response
+	queryResponses              chan queryResponse
+	statsCh                     chan QueryProgressInfo
+	usingSpooledProtocol        bool
+	allSegmentsProccess         bool
+	spooledSegmentsMetadata     chan spooledMetadata
+	spooledSegmentToDecode      chan segmentToDecode
+	decodedSegment              chan decodedSegment
+	waitSegmentDecodersWorkers  sync.WaitGroup
+	waitDownloadSegmentsWorkers sync.WaitGroup
+	startSpoolingWorkersOnce    sync.Once
+	cancelDownloadWorkers       context.CancelFunc
+	cancelDecodersWorkers       context.CancelFunc
+	spoolingProcesserDone       chan struct{}
+	errors                      chan error
+	doneCh                      chan struct{}
+}
+
+type segmentToDecode struct {
+	segmentIndex int
+	encoding     string
+	data         []byte
+	metadata     segmentMetadata
+}
+
+type decodedSegment struct {
+	rowOffset int64
+	queryData []queryData
+}
+
+func (st *driverStmt) startSpoolingProtocolWorkers(ctx context.Context) {
+	st.usingSpooledProtocol = true
+	st.spoolingProcesserDone = make(chan struct{})
+	downloadSegmentsCtx, cancelDownloadWorkers := context.WithTimeout(context.WithoutCancel(ctx), DefaultCancelQueryTimeout)
+	st.cancelDownloadWorkers = cancelDownloadWorkers
+	decodeSegmentCtx, cancelDecodersWorkers := context.WithTimeout(context.WithoutCancel(ctx), DefaultCancelQueryTimeout)
+	st.cancelDecodersWorkers = cancelDecodersWorkers
+	st.spooledSegmentsMetadata = make(chan spooledMetadata)
+	st.spooledSegmentToDecode = make(chan segmentToDecode)
+	st.startDownloadSegmentsWorkers(downloadSegmentsCtx)
+	st.startSegmentsDecodersWorkers(decodeSegmentCtx)
+}
+
+func (st *driverStmt) startDownloadSegmentsWorkers(ctx context.Context) {
+	for i := 0; i < DefaultSpoolingDownloadWorkers; i++ {
+		st.waitDownloadSegmentsWorkers.Add(1)
+		go func() {
+			defer st.waitDownloadSegmentsWorkers.Done()
+			for {
+				select {
+				case metadata := <-st.spooledSegmentsMetadata:
+
+					if metadata.encoding == "" {
+						st.cancelDownloadWorkers()
+						return
+					}
+
+					segmentFetcher := &SegmentFetcher{
+						ctx:             ctx,
+						httpClient:      st.conn.httpClient,
+						spooledMetadata: metadata,
+					}
+					segment, err := segmentFetcher.fetchSegment()
+					if err != nil {
+						st.errors <- err
+						return
+					}
+
+					st.spooledSegmentToDecode <- segmentToDecode{
+						segmentIndex: metadata.segmentIndex,
+						encoding:     metadata.encoding,
+						data:         segment,
+						metadata:     metadata.metadata,
+					}
+
+				case <-st.doneCh:
+					return
+				case <-ctx.Done():
+					return
+				}
+			}
+		}()
+	}
+}
+
+func (st *driverStmt) startSegmentsDecodersWorkers(ctx context.Context) {
+	for i := 0; i < DefaultSpoolingDecodersWorkers; i++ {
+		st.waitSegmentDecodersWorkers.Add(1)
+		go func() {
+			defer st.waitSegmentDecodersWorkers.Done()
+			for {
+				select {
+				case segmentToDecode := <-st.spooledSegmentToDecode:
+
+					if segmentToDecode.encoding == "" {
+						st.cancelDecodersWorkers()
+						return
+					}
+
+					segment, err := decodeSegment(segmentToDecode.data, segmentToDecode.encoding, segmentToDecode.metadata)
+					if err != nil {
+						st.errors <- fmt.Errorf("failed to decode spooled segment at index %d: %v", segmentToDecode.segmentIndex, err)
+						return
+					}
+
+					st.decodedSegment <- decodedSegment{
+						rowOffset: segmentToDecode.metadata.rowOffset,
+						queryData: segment,
+					}
+				case <-st.doneCh:
+					return
+				case <-ctx.Done():
+					return
+				}
+			}
+		}()
+	}
 }
 
 var (
@@ -727,9 +843,34 @@ func (st *driverStmt) Close() error {
 	}
 	for range st.httpResponses {
 	}
+
+	if st.cancelDownloadWorkers != nil {
+		st.cancelDownloadWorkers()
+	}
+
+	st.waitDownloadSegmentsWorkers.Wait()
+
+	if st.spooledSegmentsMetadata != nil {
+		close(st.spooledSegmentsMetadata)
+	}
+
+	if st.spooledSegmentToDecode != nil {
+		close(st.spooledSegmentToDecode)
+	}
+
+	if st.cancelDecodersWorkers != nil {
+		st.cancelDecodersWorkers()
+	}
+
+	st.waitSegmentDecodersWorkers.Wait()
+
 	close(st.nextURIs)
 	close(st.errors)
 	st.doneCh = nil
+	st.cancelDownloadWorkers = nil
+	st.spooledSegmentsMetadata = nil
+	st.spooledSegmentToDecode = nil
+	st.cancelDecodersWorkers = nil
 	return nil
 }
 
@@ -1020,9 +1161,11 @@ func (st *driverStmt) exec(ctx context.Context, args []driver.NamedValue) (*stmt
 
 	st.doneCh = make(chan struct{})
 	st.nextURIs = make(chan string)
+	st.decodedSegment = make(chan decodedSegment)
 	st.httpResponses = make(chan *http.Response)
 	st.queryResponses = make(chan queryResponse)
 	st.errors = make(chan error)
+
 	go func() {
 		defer close(st.httpResponses)
 		for {
@@ -1134,6 +1277,104 @@ func (st *driverStmt) exec(ctx context.Context, args []driver.NamedValue) (*stmt
 		st.conn.progressUpdaterPeriod.LastQueryState = sr.Stats.State
 	}
 	return &sr, handleResponseError(resp.StatusCode, sr.Error)
+}
+
+type SegmentFetcher struct {
+	ctx             context.Context
+	httpClient      http.Client
+	spooledMetadata spooledMetadata
+}
+
+func (sf *SegmentFetcher) roundTrip(req *http.Request) (*http.Response, error) {
+	delay := 100 * time.Millisecond
+	const maxDelayBetweenRequests = float64(15 * time.Second)
+	timer := time.NewTimer(0)
+	defer timer.Stop()
+	for {
+		select {
+		case <-timer.C:
+			resp, err := sf.httpClient.Do(req)
+			if err != nil {
+				return nil, &ErrQueryFailed{Reason: err}
+			}
+			switch resp.StatusCode {
+			case http.StatusOK:
+				return resp, nil
+			case http.StatusBadGateway, http.StatusServiceUnavailable, http.StatusGatewayTimeout:
+				resp.Body.Close()
+				timer.Reset(delay)
+				delay = time.Duration(math.Min(
+					float64(delay)*math.Phi,
+					maxDelayBetweenRequests,
+				))
+				continue
+			default:
+				return nil, newErrQueryFailedFromResponse(resp)
+			}
+		}
+	}
+}
+
+func (sf *SegmentFetcher) fetchSegment() ([]byte, error) {
+	ctx := context.WithoutCancel(sf.ctx)
+	req, err := http.NewRequestWithContext(ctx, "GET", sf.spooledMetadata.uri, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	for k, v := range sf.spooledMetadata.headers {
+		headerSlice, ok := v.([]interface{})
+		if !ok {
+			return nil, fmt.Errorf("unsupported header type %T", v)
+		}
+
+		if len(headerSlice) == 0 {
+			continue
+		}
+
+		if len(headerSlice) > 1 {
+			return nil, fmt.Errorf("multiple values for header %s", k)
+		}
+
+		header, ok := headerSlice[0].(string)
+		if !ok {
+			return nil, fmt.Errorf("unsupported header value type %T", headerSlice[0])
+		}
+		req.Header.Add(k, header)
+	}
+
+	resp, err := sf.roundTrip(req)
+	if err != nil {
+		return nil, fmt.Errorf("error fetching segment from uri '%s': %v", sf.spooledMetadata.uri, err)
+	}
+
+	data, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("error reading response body: %v", err)
+	}
+
+	//acknowledge the segment read
+	go func() {
+		// TODO: handle ack erros
+		ackReq, err := http.NewRequestWithContext(ctx, "GET", sf.spooledMetadata.ackUri, nil)
+		if err != nil {
+			return
+		}
+
+		for k, values := range req.Header {
+			for _, v := range values {
+				ackReq.Header.Add(k, v)
+			}
+		}
+
+		resp, err := sf.httpClient.Do(ackReq)
+		if err != nil {
+			return
+		}
+		resp.Body.Close()
+	}()
+
+	return data, nil
 }
 
 func formatStringLiteral(query string) string {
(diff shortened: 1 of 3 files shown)


trino/trino.go
```diff
@@ -1715,6 +1907,120 @@ func (qr *driverRows) fetch() error {
 	}
 }
 
+func (qr *driverRows) waitForAllSpoolingWorkersFinish() error {
+	spoolingDecodersWorkersDone, spoolingFetcherWorkersDone := qr.stmt.waitForSpoolingWorkers()
```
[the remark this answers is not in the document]
Flgado: This is not blocking. In this moment we not have more nextUri's to proccess but the workers are still doing some job. The next loop will notify workers that they not have more work to do. So the last worker will get the empty struct(maybe there is a better way yes) and inform all others that they should stop. This is to make sure that all segments are already process before I start closing the channels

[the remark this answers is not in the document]
nineinchnick: Right, I read `waitForSpoolingWorkers` wrong. Maybe you could inline it here, but let's address other comments first.

trino/trino.go
```diff
@@ -153,6 +157,9 @@ const (
 
 	mapKeySeparator   = ":"
 	mapEntrySeparator = ";"
+
+	SpoolingDownloadWorkers = "SpoolingDownloadWorkers"
+	SpoolingDecodersWorkers = "SpoolingDecodersWorkers"
```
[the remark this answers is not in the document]
Flgado: As we will choose a high default value I was thinking that maybe some users want to use less workers to avoid unnecessary overhead. Let me know your opinion, if not make sense I can remove it 👍🏻

trino/trino.go
```diff
@@ -690,15 +697,140 @@ func newErrQueryFailedFromResponse(resp *http.Response) *ErrQueryFailed {
...
+	st.usingSpooledProtocol = true
+	if st.numberOfDecodersWorkers == 0 {
+		st.numberOfDecodersWorkers = DefaultSpoolingDecodersWorkers
+	}
+
+	if st.numberOfDownloadWorkers == 0 {
+		st.numberOfDownloadWorkers = DefaultSpoolingDownloadWorkers
+	}
+
+	downloadSegmentsCtx, cancelDownloadWorkers := context.WithTimeout(context.WithoutCancel(ctx), DefaultCancelQueryTimeout)
```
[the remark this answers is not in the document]
Flgado: Should we not have a timeout for downloading the segments ? Is better right ?

trino/trino.go
```diff
@@ -727,9 +859,40 @@ func (st *driverStmt) Close() error {
...
+	}
+
+	st.waitSegmentDecodersWorkers.Wait()
+
 	close(st.nextURIs)
 	close(st.errors)
+
 	st.doneCh = nil
+	st.cancelDownloadWorkers = nil
+	st.spooledSegmentsMetadata = nil
```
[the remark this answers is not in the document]
Flgado: One question, why we need to set the channels to nil ? The driverStmt is not created each time we query trino with this channels as nil ?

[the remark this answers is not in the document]
nineinchnick: Technically, we don't have to do this, but this might help detect some issues if any part of the code tries using this channel after the statement is closed. We'll get a different error than when trying to read from a closed channel, and we'll know it's not an issue in the `Close()` method, but a "use-after-free" user error.

trino/trino.go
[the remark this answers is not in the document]
nineinchnick: I asked why we need to control the number of downloader and decoder workers separately, not why we need to control them at all.

trino/trino.go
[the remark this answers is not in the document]
nineinchnick: We don't need a separate timeout for every segment, _unless_ we can retry downloading them.

trino/trino.go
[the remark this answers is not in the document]
Flgado: ah sorry I not read it right 😄 . Agree maybe make sense to have a single value to control both of then. If someone have a use case where they need to be separated we can easily change after. At least I am not seeing any. Will add that capacity again after the review 👍🏻 
Thanks !

trino/trino.go
```diff
@@ -1136,6 +1308,104 @@ func (st *driverStmt) exec(ctx context.Context, args []driver.NamedValue) (*stmt
...
+				))
+				continue
+			default:
+				return nil, newErrQueryFailedFromResponse(resp)
+			}
+		}
+	}
+}
+
+func (sf *SegmentFetcher) fetchSegment() ([]byte, error) {
```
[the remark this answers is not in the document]
Flgado: I needed to move the method to a worker :/

trino/trino.go
```diff
@@ -1715,6 +1822,205 @@ func (qr *driverRows) fetch() error {
...
 
+func (st *driverStmt) startSpoolingProtocolWorkers(ctx context.Context) {
+	st.usingSpooledProtocol = true
+
+	downloadSegmentsCtx, cancelDownloadWorkers := context.WithTimeout(context.WithoutCancel(ctx), DefaultCancelQueryTimeout)
+	st.cancelDownloadWorkers = cancelDownloadWorkers
+	decodeSegmentCtx, cancelDecodersWorkers := context.WithTimeout(context.WithoutCancel(ctx), DefaultCancelQueryTimeout)
+	st.cancelDecodersWorkers = cancelDecodersWorkers
+
+	st.spooledSegmentsMetadata = make(chan spooledMetadata, 100)
```
nineinchnick, line 1833: How did you choose the size here? We need to document this well.

We want to make this as big as possible, so the coordinator can finish queries and free up resources, before the client receives all the data. 

Assuming the default max segment size of 16MB, queueing up to 100 segments means only queries returning up to about 1.5GB of data will be finished quickly, before the client downloads all the data.

If downloading is slow, we'll be blocking on writing to this channel. If we block for too long, the query will time out on the coordinator. We can send heartbeats as HEAD requests. Before we implement that, we can keep the size of this channel constrained.

We should also do some estimates on memory usage when this channel is full. I think a practical limit would be the default max size of a single segment, so 16MB.

trino/trino.go
```diff
@@ -1642,14 +1818,44 @@ func handleResponseError(status int, respErr ErrTrino) error {
...
+				}
+
+				requiredSize := int(segment.rowOffset + int64(len(segment.queryData)))
+				if len(qr.data) < requiredSize {
+					newData := make([]queryData, requiredSize)
+					copy(newData, qr.data)
+					qr.data = newData
+				}
+				for i, row := range segment.queryData {
+					qr.data[int(segment.rowOffset)+i] = row
```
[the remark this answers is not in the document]
nineinchnick: How many segments were used? 20M rows might compress well, depending on the number of columns, their data types, and the actual values.

What was the memory usage in your test?

trino/trino.go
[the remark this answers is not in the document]
nineinchnick: Generally, structs and methods should be defined after their first use, so here is not the best place anyway. But it's a nitpick

trino/trino.go
```diff
@@ -1642,6 +1708,62 @@ func handleResponseError(status int, respErr ErrTrino) error {
...
+		defer close(qr.stmt.spoolingRowsChannel)
+		defer close(qr.stmt.spoolingProcesserDone)
+
+		buffer := make(map[int64][]queryData)
+		var nextExpectedOffset int64 = 0
+
+		for {
+			select {
+			case segment, ok := <-qr.stmt.decodedSegment:
+				if !ok {
```
nineinchnick, line 1723: When this can happen, and why is this necessary? You still reference `segment.queryData` after this block.

trino/trino.go
```diff
@@ -1245,7 +1410,34 @@ func (qr *driverRows) Next(dest []driver.Value) error {
 			qr.err = err
 			return err
 		}
+	} else if qr.stmt.usingSpooledProtocol && (qr.rowindex >= len(qr.data) || qr.data == nil) {
+		var ok bool
+		select {
+		case qr.data, ok = <-qr.stmt.spoolingRowsChannel:
```
nineinchnick, line 1416: This channel is only initialized after calling `fetch()`, but I don't see it being called here.

trino/trino.go
```diff
@@ -1642,6 +1708,62 @@ func handleResponseError(status int, respErr ErrTrino) error {
 	}
 }
 
+func (qr *driverRows) startSegmentProcessor() {
+	qr.stmt.spoolingRowsChannel = make(chan []queryData)
```
nineinchnick, line 1712: Why you initialize the channel here, and not earlier, for example, together with `qr.stmt.spoolingProcesserDone`?

trino/trino.go
```diff
@@ -1642,6 +1708,62 @@ func handleResponseError(status int, respErr ErrTrino) error {
...
+							return
+						}
+
+						delete(buffer, nextExpectedOffset)
+
+						nextExpectedOffset += int64(len(data))
+					}
+				}
+
+				buffer[segment.rowOffset] = segment.queryData
```
nineinchnick, line 1742: Are we going to allocate the buffer for all the rows in query results? This needs to have a finite upper bound.

trino/trino.go
```diff
@@ -1715,6 +1822,205 @@ func (qr *driverRows) fetch() error {
...
+	st.waitDownloadSegmentsWorkers.Add(DefaultSpoolingDownloadWorkers)
+	for i := 0; i < DefaultSpoolingDownloadWorkers; i++ {
+		go func() {
+			defer st.waitDownloadSegmentsWorkers.Done()
+			for {
+				select {
+				case metadata, ok := <-st.spooledSegmentsMetadata:
+
+					if !ok {
+						st.cancelDownloadWorkers()
```
nineinchnick, line 1852: Why is this needed? All workers should exit when `spooledSegmentsMetadata` is closed.

trino/trino.go
Flgado: my bad, should be with the others channel

Flgado:
image

To better understand and talk about what is going on with the go orchestration
@nineinchnick

trino/trino.go
Flgado: sorry didn't understood. This only be use when spooling protocol is enabled. My idea is in the future for the direct protocol instead of calling always the fetch method we can also use this channel

nineinchnick: The flow is not very straightforward, so I'd appreciate a code comment here to explain when this channel gets initiated.

When we call `fetch()` to get the first query response, it notices the new `data` structure with segments, and calls `startSpoolingProtocolWorkers` that sets `usingSpooledProtocol` to true and initializes the `spoolingRowsChannel` channel.

trino/trino.go
nineinchnick: We should be using relative offsets, not absolute offsets here, to avoid growing the buffer too much.

trino/trino.go
```diff
@@ -1642,6 +1730,62 @@ func handleResponseError(status int, respErr ErrTrino) error {
...
+					}
+
+					select {
+					case qr.stmt.spoolingRowsChannel <- data:
+					case <-qr.doneCh:
+						return
+					}
+
+					delete(buffer, nextExpectedOffset)
+					nextExpectedOffset += int64(len(data))
```
nineinchnick, line 1774: We need a way to detect when this is going to get stuck because of a server error, when it'll send data with a gap. It should be possible to check that all downloaders are blocked, all decoders finished, so no more segments will arrive here, and we still don't have the next expected offset.

We need a test that will simulate this.

trino/trino.go
```diff
@@ -1715,6 +1844,225 @@ func (qr *driverRows) fetch() error {
...
+				qr.rowsAffected = qresp.UpdateCount
+				qr.scheduleProgressUpdate(qresp.ID, qresp.Stats)
+			}
+		}
+	}()
+}
+
+func (qr *driverRows) waitForAllSpoolingWorkersFinish() {
+	close(qr.stmt.spooledSegmentsMetadata)
+	qr.stmt.waitDownloadSegmentsWorkers.Wait()
```
nineinchnick, line 2000: Why do we need to wait here?

trino/trino.go
```diff
@@ -1715,6 +1844,225 @@ func (qr *driverRows) fetch() error {
...
+				select {
+				case metadata, ok := <-st.spooledSegmentsMetadata:
+
+					if !ok {
+						return
+					}
+
+					// downloadPermitCh is a semaphore channel used to limit the number of concurrent
+					// segment downloads. Each worker must acquire a permit by sending an empty struct{}
+					// to the channel before starting a download, ensuring memory usage stays within controlled bounds.
```
nineinchnick, line 1889:
```suggestion
					// downloadPermitCh blocks if there are too many out-of-order segments.
					// Once all currently downloaded segments are downloaded, decoded,
					// and can be ordered, this channel will be drained.
```

trino/trino.go
```diff
@@ -1642,6 +1730,62 @@ func handleResponseError(status int, respErr ErrTrino) error {
...
+
+					delete(buffer, nextExpectedOffset)
+					nextExpectedOffset += int64(len(data))
+
+					if backlogCount > 0 {
+						backlogCount--
+					}
+				}
+
+			case <-downloadPermitCh:
```
nineinchnick, line 1781: Do this in a separate `select` statement. Read from this channel once for every segment sent to `spoolingRowsChannel`.

trino/trino.go
```diff
@@ -1642,6 +1730,62 @@ func handleResponseError(status int, respErr ErrTrino) error {
...
+		defer close(qr.stmt.spoolingProcesserDone)
+		defer close(qr.stmt.downloadPermitCh)
+		buffer := make(map[int64][]queryData)
+		var nextExpectedOffset int64 = 0
+		maxBacklog := 2 * qr.stmt.spoolingMaxOutOfOrderSegments
+		backlogCount := 0
+
+		for {
+			var downloadPermitCh <-chan struct{}
+			if backlogCount <= maxBacklog {
```
nineinchnick, line 1745: You should not need to check this explicitly, the size of the `downloadPermitCh` should control this - it will block downloaders when it gets full.

trino/trino.go
```diff
@@ -1136,6 +1225,104 @@ func (st *driverStmt) exec(ctx context.Context, args []driver.NamedValue) (*stmt
...
+	httpClient      http.Client
+	spooledMetadata spooledMetadata
+}
+
+func (sf *SegmentFetcher) roundTrip(req *http.Request) (*http.Response, error) {
+	delay := 100 * time.Millisecond
+	const maxDelayBetweenRequests = float64(15 * time.Second)
+	timer := time.NewTimer(0)
+	defer timer.Stop()
+	for {
```
nineinchnick, line 1239: There should be an upper bound for the number of retries.

trino/trino.go
```diff
@@ -1136,6 +1225,104 @@ func (st *driverStmt) exec(ctx context.Context, args []driver.NamedValue) (*stmt
...
+		select {
+		case <-timer.C:
+			resp, err := sf.httpClient.Do(req)
+			if err != nil {
```
nineinchnick, line 1243: Can we retry on network timeouts?

trino/trino.go
```diff
@@ -1715,6 +1844,225 @@ func (qr *driverRows) fetch() error {
...
+					segmentFetcher := &SegmentFetcher{
+						ctx:             ctx,
+						httpClient:      st.conn.httpClient,
+						spooledMetadata: metadata,
+					}
+
+					segment, err := segmentFetcher.fetchSegment()
+					if err != nil {
+						st.errors <- err
+						st.cancelDownloadWorkers()
```
nineinchnick, line 1907: Cancelling of any remaining work should happen in the error handling function.

trino/trino.go
```diff
@@ -1715,6 +1844,225 @@ func (qr *driverRows) fetch() error {
...
+					return
+				case <-ctx.Done():
+					return
+				}
+			}
+		}()
+	}
+}
+
+func (qr *driverRows) proccessSpollingSegments() {
```
nineinchnick, line 1962: Why do we need this method? It duplicates a large part of `fetch()`

trino/trino.go
Flgado: hmm and what about the first state where no data arrive yet to "F"

Flgado: Ah saw your comment now about controlling the downloads with the size of the channel. Sure will change it

trino/trino.go
Flgado: hmm but using the map we already not avoid that?
The map have the started offset of the slice, so as soon as we have any data in order he delete the key and the respective slice

nineinchnick: I don't think a map is the best data structure here, but I might be wrong. I'll do some benchmarks first.

trino/trino.go
nineinchnick: I don't remember exactly, but probably because there's a timeout in the context. The difference is that iterating over nextUri cannot be parallelized, so a single timeout is ok. We need a shorter timeout here and handle it's expiration.

trino/trino.go
nineinchnick: You contradicted yourself here. If this can never happen, remove this condition.

Flgado: wait, you are talking about the if !ok condition ?
I remove the logic that was inside the condition. that could never happen, now is just returning and finish the goroutine. Inside this I was cleaning the buffer but that didn't make sense because if the channel is close is suppose that we not have more data in the buffer. That was not what you are referring to ?

trino/trino.go
nineinchnick: What I was suggesting is just more explicit. If we'd use a slice here, we'd have to sort it. But it would only require a single allocation, since it would have a constant size. A map might have a good enough implementation for this. A benchmark is the only way to be sure.

trino/trino.go
Flgado: Was only to avoid extra work as soon as possible. It can take some time until error is handled and cancel all workers. But yeah make sense, make thinks more organized, will remove it

trino/trino.go
Flgado: The fetch function is synchronous, while this one is asynchronous and is only used to handle the spooling protocol. My idea for the future is to also make the direct protocol asynchronous and share a channel with the next() method. This way, we can have a single function for both protocols, and a simple synchronous function whose only responsibility is to start the goroutine that processes the decoded data coming from Trino.

However, I believe achieving this will require significant refactoring of the next() and fetch functions in this PR.

What do you think?

README.md
```diff
@@ -341,6 +341,122 @@ following types:
...
+  - Higher values allow more segments to be downloaded in parallel, increasing memory usage.
+  - Lower values reduce memory usage but may limit parallelism and throughput.
+  - If your application or environment is memory-constrained, consider lowering this value.
+
+#### Example: Customizing Spooling Parameters
+
+```go
+rows, err := db.Query(
+    query,
+    sql.Named("encoding", "json+zstd"),
```
Flgado, line 390: @nineinchnick  maybe we can have this parameters as exported "enums" so users can use for example `trinoEncoding` instead of "encoding" ? wyt ?

trino/trino.go
nineinchnick: We need to handle a case where the server returns out-of-order data. If we know we can issue a maximum number of N download permits, it means there can only be N-1 out of order segments. If the last segment received here still doesn't have the expected offset, we should fail the whole query, instead of waiting for it to time out.

README.md
```diff
@@ -341,6 +341,122 @@ following types:
 > `VARBINARY` columns are returned as base64-encoded strings when used within
 > `ROW`, `MAP`, or `ARRAY` values.
 
+## Spooling Protocol
+The client spooling protocol requires a [Trino server with spooling protocol support](https://trino.io/docs/current/client/client-protocol.html#spooling-protocol).
+
+### Enabling Spooling
+
+Enabled the spooling protocol by specifying a supported encoding in the `encoding` parameter:
```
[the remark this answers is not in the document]
Flgado: yup! done

trino/trino.go
```diff
@@ -1642,6 +1775,67 @@ func handleResponseError(status int, respErr ErrTrino) error {
...
+		for {
+			select {
+			case segment, ok := <-qr.stmt.decodedSegment:
+				if !ok {
+					return
+				}
+
+				buffer = append(buffer, segment)
+
+				if nextExpectedOffset == segment.rowOffset {
```
nineinchnick, line 1796: Nit: reverse this condition and add a `continue`.

trino/trino.go
nineinchnick: It's enough to check if the buffer is full. It has a fixed size and should never grow. If this happens, it means the server is returning incorrect data, and we should terminate with an error.

trino/trino.go
```diff
@@ -690,15 +705,44 @@ func newErrQueryFailedFromResponse(resp *http.Response) *ErrQueryFailed {
...
+	nextURIs                      chan string
+	httpResponses                 chan *http.Response
+	queryResponses                chan queryResponse
+	statsCh                       chan QueryProgressInfo
+	usingSpooledProtocol          bool
+	spoolingMaxOutOfOrderSegments int
+	spoolingWorkerCount           int
+	spooledSegmentsMetadata       chan spooledMetadata
+	spooledSegmentToDecode        chan segmentToDecode
+	decodedSegment                chan decodedSegment
```
nineinchnick, line 720: Nit: use the plural form for all channels; it would be more consistent with existing channels and help distinguish channel names from struct names.
```suggestion
	decodedSegments                chan decodedSegment
```

trino/trino.go
```diff
@@ -1715,6 +1894,272 @@ func (qr *driverRows) fetch() error {
...
+		defer close(st.rateLimitDoneCh)
+		defer close(st.segmentThrottleCh)
+		for {
+			select {
+			case segmentToProccess, ok := <-st.segmentsToProccess:
+				if !ok {
+					return
+				}
+
+				// downloadPermitCh blocks if there are too many out-of-order segments.
```
nineinchnick, line 1938: Update the comment after renaming the channel name

trino/trino_test.go
```diff
@@ -1286,6 +1286,302 @@ func TestSpoolingProtocolSpooledSegmentDecoders(t *testing.T) {
 	}
 }
 
+func TestSpoolingProtocolOutOfOrderSegmentDownload(t *testing.T) {
+	testcases := []struct {
+		Name              string
+		Segments          []map[string]interface{}
+		ExpectedResult    []int
+		Encoding          string
+		FirstSegmentData  []byte
```
nineinchnick, line 1295: These seem unnecessary; remove them and construct the data on the fly.

trino/trino_test.go
```diff
@@ -1286,6 +1286,302 @@ func TestSpoolingProtocolSpooledSegmentDecoders(t *testing.T) {
 	}
 }
 
+func TestSpoolingProtocolOutOfOrderSegmentDownload(t *testing.T) {
+	testcases := []struct {
```
nineinchnick, line 1290: Are we ever going to have more cases in this test? If not, remove the loop

trino/trino_test.go
```diff
@@ -1286,6 +1286,302 @@ func TestSpoolingProtocolSpooledSegmentDecoders(t *testing.T) {
...
+						},
+						Data: map[string]interface{}{
+							"encoding": tc.Encoding,
+							"segments": tc.Segments,
+						},
+					})
+					return
+				}
+				if r.URL.Path == "/v1/spooled/download/jKaLK0aVkNp2ixl6BOuwGMJ0nRjbUVKLHW_f3-I-1Cc=" {
+					time.Sleep(3 * time.Second)
```
nineinchnick, line 1368: This is a bad way of testing concurrency. Create 2 empty struct channels, block on reading from them, and close them in the right order.

trino/trino.go
```diff
@@ -1715,6 +1894,272 @@ func (qr *driverRows) fetch() error {
...
+	st.cancelDownloadWorkers = cancelDownloadWorkers
+	decodeSegmentCtx, cancelDecodersWorkers := context.WithCancel(context.WithoutCancel(ctx))
+	st.cancelDecodersWorkers = cancelDecodersWorkers
+
+	st.segmentsToProccess = make(chan segmentToProccess, 1000)
+	st.spooledSegmentsMetadata = make(chan spooledMetadata, st.spoolingMaxOutOfOrderSegments)
+	st.spooledSegmentToDecode = make(chan segmentToDecode, st.spoolingMaxOutOfOrderSegments)
+	st.spoolingRowsChannel = make(chan []queryData)
+	st.spoolingProcesserDone = make(chan struct{})
+	st.rateLimitDoneCh = make(chan struct{})
```
nineinchnick, line 1918: This name is not accurate; it's used to wait for the segment dispatcher to finish.
```suggestion
	st.segmentDispatcherDoneCh = make(chan struct{})
```

trino/trino.go
```diff
@@ -1290,89 +1555,50 @@ type queryResponse struct {
...
+		return spooledMetadata{}, fmt.Errorf("missing or invalid 'headers' field in spooled segment at index %d", segmentIndex)
 	}
 
-	return queryData, nil
+	return result, nil
 }
 
-func parseSpoolingMetadata(metadata map[string]interface{}) (spoolingMetadata, error) {
-	result := spoolingMetadata{
+func parseSpoolingMetadata(metadata map[string]interface{}) (segmentMetadata, error) {
```
nineinchnick, line 1600: Rename this to `parseSegmentMetadata`, The name `parseSpoolingMetadata` is confusing with `parseSpooledMetadata`.

README.md
```diff
@@ -341,6 +341,69 @@ following types:
...
+  `sql.Named("max_out_of_order_segments", "N")`  
+  Sets the maximum number of segments that can be downloaded and buffered out-of-order before blocking further downloads.  
+  **Default:** `10`  
+  **Considerations:**  
+  - Higher values increase the potential memory usage, but actual usage depends on download behavior and may be lower in practice.
+  - Higher values reduce the chance that one slow or stalled segment will block the download of additional segments.
+  - Lower values reduce memory usage but may limit parallelism and throughput.
+
+> **Note:**  
+> It does not make sense to set `spooling_worker_count` higher than `max_out_of_order_segments`. For example, if you allow 10 out-of-order segments but have 50 workers, most workers will be blocked waiting for a slot. For best efficiency, keep `spooling_worker_count` less than or equal to `max_out_of_order_segments`.
```
nineinchnick, line 386: If it doesn't make sense, we should not allow it - throw an error.

README.md
```diff
@@ -341,6 +341,69 @@ following types:
...
+```go
+rows, err := db.Query(
+    query,
+    sql.Named("encoding", "json+zstd"),
+    sql.Named("spooling_worker_count", "8"),
+    sql.Named("max_out_of_order_segments", "20"),
+)
+```
+
+### Tuning Guidelines
```
nineinchnick, line 399: This duplicates the `Configuration Options` section

nineinchnick (approved):
A few nitpicks, but this is ready to be merged. Thanks for working on this and your patience with reviews!

trino/trino.go
```diff
@@ -1136,6 +1254,123 @@ func (st *driverStmt) exec(ctx context.Context, args []driver.NamedValue) (*stmt
...
+
+			default:
+				return nil, newErrQueryFailedFromResponse(resp)
+			}
+		}
+	}
+}
+
+func (sf *SegmentFetcher) fetchSegment() ([]byte, error) {
+	ctx := context.WithoutCancel(sf.ctx)
```
nineinchnick, line 1313: Why are you removing the cancel from the context? If the connection gets stuck, it won't be possible to cancel the query.

nineinchnick (approved):
Please squash all commits

trino/trino.go
Flgado: Nice catch! This was here because earlier I wasn’t always handling context cancellation correctly, and I was canceling the download the wrong way. But now we’re doing it the right way. Either way, this is wrong thanks! 😊

trino/trino_test.go
```diff
@@ -1299,6 +1299,397 @@ func TestSpoolingProtocolSpooledSegmentDecoders(t *testing.T) {
...
+		case "/v1/spooled/download/jKaLK0aVkNp2ixl6BOuwGMJ0nRjbUVKLHW_f3-I-1Cc=":
+			<-doneSecond
+			w.Write([]byte("[[1000]]"))
+			return
+
+		case "/v1/spooled/download/jKaLK0aVkNp2ixl6BOuwGMJ0nRjbUVKLHW_f3-I-1Cc1=":
+			<-doneFirst
+			w.Write([]byte("[[1001]]"))
+			go func() {
+				time.Sleep(2 * time.Second)
```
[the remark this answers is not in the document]
Flgado: I think it make sense yes, because we only are testing that the last worker kept things in order if they came out of order. I will follow that approach 👍🏻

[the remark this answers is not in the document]
Flgado: check it now @nineinchnick  :)

nineinchnick:
@Flgado, please rebase to solve conflicts and squash all commits.

PR #142, opened 2025-05-08
Add hstore support

JorgeMoraisDatadog:
Adds hstore as new type


nineinchnick:
That's not a Trino type, but Postgresql

PR #144, opened 2025-05-15
Add support for catalog roles

Flgado:
This PR adds support for specifying catalog roles in the Trino Go client via both the DSN string and the `Config` struct. Roles are passed as either:

- A `map[string]string` mapping catalogs to roles (e.g. `{"hive": "analyst", "system": "admin"}`)
- A string (e.g. `"admin"`), which is treated as a shorthand for `system=admin`

The roles are URL-encoded and appended to the DSN as a semicolon-separated list of `catalog=role` entries.

### Example

```go
// Using role map
c := &Config{
    ServerURI: "https://user@localhost:8080",
    Roles: map[string]string{"hive": "analyst", "system": "admin"},
}

// Using string for system role
c := &Config{
    ServerURI: "https://user@localhost:8080",
    Roles: "admin", // Equivalent to system=admin
}

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,6 +249,46 @@ Default:        nil
 
 The `queryTimeout` parameter sets a timeout for the query. If the query takes longer than the timeout, it will be cancelled. If it is not set the default context timeout will be used.
 
+
+#### `roles`
+
+```
+Type:           string
+Format:         roles=catalog1=role1;catalog2=role2
+Valid values:   semicolon-separated list of catalog-to-role assignments
+Default:        empty
+```
+The roles parameter defines authorization roles to assume for one or more catalogs during the Trino session.
+
+You can assign roles either as a map of catalog-to-role pairs or a string. When a string is used, it applies the role to the `system` catalog by default.
+
+##### Example
+
+``` go
+c := &Config{
+	ServerURI:         "https://foobar@localhost:8090",
+	SessionProperties: map[string]string{"query_priority": "1"},
+	Roles:             map[string]string{"catalog1": "role1", "catalog2": "role2"},
+}
+
+dsn, err := c.FormatDSN()
+// Result: https://foobar@localhost:8090?roles=catalog1%3Drole1%3Bcatalog2%3Drole2&session_properties=query_priority%3A1
+```
+
+**Example using a string (applies to system catalog)**
+
+``` go
+c := &Config{
+	ServerURI:         "https://foobar@localhost:8090",
+	SessionProperties: map[string]string{"query_priority": "1"},
+	Roles:             "admin", // equivalent to map[string]string{"system": "admin"}
+}
+
+dsn, err := c.FormatDSN()
+// Result: https://foobar@localhost:8090?roles=system%3Dadmin&session_properties=query_priority%3A1
+```
+
+
 #### Examples
 
 ```
--- a/trino/trino.go
+++ b/trino/trino.go
@@ -126,6 +126,7 @@ const (
 	trinoSetSessionHeader      = trinoHeaderPrefix + `Set-Session`
 	trinoClearSessionHeader    = trinoHeaderPrefix + `Clear-Session`
 	trinoSetRoleHeader         = trinoHeaderPrefix + `Set-Role`
+	trinoRoleHeader            = trinoHeaderPrefix + `Role`
 	trinoExtraCredentialHeader = trinoHeaderPrefix + `Extra-Credential`
 
 	trinoProgressCallbackParam       = trinoHeaderPrefix + `Progress-Callback`
@@ -153,6 +154,8 @@ const (
 
 	mapKeySeparator   = ":"
 	mapEntrySeparator = ";"
+	mapRolesSeparator = "="
+	sistemRole        = "system"
 )
 
 var (
@@ -194,6 +197,7 @@ type Config struct {
 	AccessToken                string            // An access token (JWT) for authentication (optional)
 	ForwardAuthorizationHeader bool              // Allow forwarding the `accessToken` named query parameter in the authorization header, overwriting the `AccessToken` option, if set (optional)
 	QueryTimeout               *time.Duration    // Configurable timeout for query (optional)
+	Roles                      interface{}       // Roles (optional)
 }
 
 // FormatDSN returns a DSN string from the configuration.
@@ -214,6 +218,20 @@ func (c *Config) FormatDSN() (string, error) {
 			credkv = append(credkv, k+mapKeySeparator+v)
 		}
 	}
+
+	var roles []string
+	if c.Roles != nil {
+		if v, ok := c.Roles.(string); ok {
+			roles = append(roles, sistemRole+mapRolesSeparator+v)
+		} else if v, ok := c.Roles.(map[string]string); ok {
+			for k, v := range v {
+				roles = append(roles, k+mapRolesSeparator+v)
+			}
+		} else {
+			return "", fmt.Errorf("Invalid roles type %T", c.Roles)
+		}
+	}
+
 	source := c.Source
 	if source == "" {
 		source = "trino-go-client"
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: README.md [hunk 1]
```diff
@@ -249,6 +249,46 @@ Default:        nil
 
 The `queryTimeout` parameter sets a timeout for the query. If the query takes longer than the timeout, it will be cancelled. If it is not set the default context timeout will be used.
 
+
+#### `roles`
+
+```
+Type:           string
+Format:         roles=catalog1=role1;catalog2=role2
```
nineinchnick, line 257: The JDBC driver uses colons instead of equal signs for the separator between catalog and role. We should use the same format here. I fixed that last year for session properties in https://github.com/trinodb/trino-go-client/pull/127

Also note this should come up in integration tests, that are missing.

nineinchnick (changes requested):

README.md
Flgado: It seems that the format Trino expects is: `catalog=ROLE{role}`,.

The Memory and TPC-H connectors do not support roles. I set up Iceberg with Postgres, but it looks like that connector also doesn't support creating roles via SQL — I consistently get the error: "this connector does not support roles."

I also tried using file-based access control. While that does successfully block access to the catalog, I haven’t found a way to grant access to a specific role for the Iceberg catalog 😢.

I’ve added integration tests to verify Trino’s expected format for the x-role header. However, I’m not sure if that’s sufficient, since role behavior seems to vary quite a bit depending on the connector in use.

Any ideas or suggestions for configuring this properly would be greatly appreciated!

nineinchnick: A simple test is enough, we're not testing roles support in connectors, just things like correct syntax - the issue you found is a good example why we need tests like this.

README.md
```diff
@@ -249,6 +249,46 @@ Default:        nil
 
 The `queryTimeout` parameter sets a timeout for the query. If the query takes longer than the timeout, it will be cancelled. If it is not set the default context timeout will be used.
 
+
+#### `roles`
+
+```
+Type:           string
+Format:         roles=catalog1=ROLE{role1},catalog2=ROLE{role2}
```
nineinchnick, line 257: Let's not expose the internal syntax to users. This was done in the Python driver, but now is considered a deprecated format. The JDBC driver should be our reference implementation, and it uses the `roles=catalog1:role1;catalog2:role2` syntax. This should match the syntax used for session properties, so you can reuse all the constants for separators, etc.

trino/trino.go
```diff
@@ -214,6 +219,20 @@ func (c *Config) FormatDSN() (string, error) {
...
 		}
 	}
+
+	var roles []string
+	if c.Roles != nil {
+		if v, ok := c.Roles.(string); ok {
+			roles = append(roles, sistemRole+mapRolesSeparator+fmt.Sprintf("ROLE{%q}", v))
+		} else if v, ok := c.Roles.(map[string]string); ok {
+			for k, v := range v {
+				roles = append(roles, k+mapRolesSeparator+fmt.Sprintf("ROLE{%q}", v))
```
nineinchnick, line 229: Don't mix string concatenation and format strings. We also don't really need constants for the system role and the separator here, if this is the only place where they're used.
```suggestion
			roles = append(roles, fmt.Sprintf("system=ROLE{%q}", v))
		} else if v, ok := c.Roles.(map[string]string); ok {
			for k, v := range v {
				roles = append(roles, fmt.Sprintf("%s=ROLE{%q}", k, v))
```

trino/trino.go
```diff
@@ -194,6 +198,7 @@ type Config struct {
 	AccessToken                string            // An access token (JWT) for authentication (optional)
 	ForwardAuthorizationHeader bool              // Allow forwarding the `accessToken` named query parameter in the authorization header, overwriting the `AccessToken` option, if set (optional)
 	QueryTimeout               *time.Duration    // Configurable timeout for query (optional)
+	Roles                      interface{}       // Roles (optional)
```
nineinchnick, line 201: Don't use `interface{}` if not necessary. We should normalize the roles string from the URL when building this struct, so it's always a map. When users build this struct directly, it's actually better to have only one way of defining roles, and being explicit about the system role.

Flgado: Sure, I’ll change it. I was mimicking the Python client in this case — they support both strings and maps. I agree it doesn’t make much sense have a specific case only for the system catalog, but I wanted to hear your opinion. Thanks! 🙂

trino/trino.go
```diff
@@ -390,6 +403,26 @@ func newConn(dsn string) (*Conn, error) {
 		queryTimeout = &d
 	}
 
+	var formatedRoles string
+	if rolesStr := query.Get("roles"); rolesStr != "" {
```
nineinchnick, line 407: Why would we allow overriding roles in query parameters, instead of setting them for the whole connection? I think this should be only possible using SQL statements, and we actually should handle the `X-Set-Role` header coming back from the server (and remove it from `unsupportedResponseHeaders`).

trino/trino.go
```diff
@@ -284,6 +295,7 @@ func (c *Config) FormatDSN() (string, error) {
 		"extra_credentials":  strings.Join(credkv, mapEntrySeparator),
 		"custom_client":      c.CustomClientName,
 		accessTokenConfig:    c.AccessToken,
+		"roles":              strings.Join(roles, mapCommaSeparator),
```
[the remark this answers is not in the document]
Flgado: No, if we send it with this format `catalog=ROLE{role};`  we get a 400. Trino is waiting comma in this case

trino/trino.go
Flgado: Yes, if we want to support it via DNS and use the JDBC format, we’ll need to find a way to convert it to the correct format. So, are you suggesting it should only be sent as an SQL statement? I’ll make the changes. 👍🏻
I also going to take a look to `X-Set-Role` thank you !

Flgado:
@nineinchnick, do we want to support the X-Trino-Role header when it's provided as a named argument?

For example: db.Query("SELECT * FROM foobar WHERE id = ?", 1, sql.Named("X-Trino-Role", ...))

If so, that should overwrite the roles on the connection string ?
Can this be a security concern ?

trino/trino.go
```diff
@@ -1187,6 +1247,15 @@ func (st *driverStmt) exec(ctx context.Context, args []driver.NamedValue) (*stmt
 					st.user = headerValue
 				}
 
+				if arg.Name == trinoRoleHeader {
+					formattedRoles, err := parseAndFormatRoles(headerValue)
```
nineinchnick: We should parse it the same way as in `ParseDSN`, but I think it's actually better to require a map being passed here, and use `formatRolesFromMap()`. If we allow strings, it's not clear which format to use.

trino/integration_test.go
```diff
@@ -1030,6 +1031,79 @@ func TestIntegrationNoResults(t *testing.T) {
...
+				dns = tt.rawDSN
+			} else {
+				dns, err = tt.config.FormatDSN()
+				if err != nil {
+					t.Fatal(err)
+				}
+			}
+
+			db := integrationOpen(t, dns)
+			_, err = db.Query("SELECT 1")
```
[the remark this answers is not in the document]
nineinchnick: 372 is very old, we can drop support for it. If we have to choose a support policy now, I'd say anything older than 2 years doesn't have to be supported explicitly - might still work, but we don't test it.

trino/integration_test.go
```diff
@@ -217,6 +228,35 @@ func TestMain(m *testing.M) {
...
+		"trino",
+		"--user", "admin",
+		"--execute", grantSQL,
+	}
+	exec, err := pool.Client.CreateExec(docker.CreateExecOptions{
+		Container: trinoResource.Container.ID,
+		Cmd:       execCmd,
+	})
+	if err != nil {
+		log.Printf("Warning: Failed to create exec for GRANT: %s", err)
```
nineinchnick: Do an early return to avoid the indentation for the `else` block

PR #167, opened 2025-11-13
handle NULL VARBINARY values correctly in ConvertValue

### LINKED ISSUE #166: NULL VARBINARY values return empty []byte instead of nil
The Trino Go client incorrectly handles NULL VARBINARY values by returning empty byte arrays ([]byte{}) instead of nil. This breaks null handling consistency and causes integration issues with downstream libraries expecting proper null values.

Mandukhai-Alimaa:
NULL VARBINARY values were incorrectly being converted to empty []byte{} instead of nil.

The root cause was that scanNullBytes() returned []byte directly, which when combined with Go's driver.Value interface conversion, lost the NULL semantics in certain scanning scenarios. (The issue was []byte(nil) getting converted to []byte{} through interface{} conversions.)

This change implements a proper NullBinary type following the same pattern as sql.NullString, sql.NullInt64, etc., ensuring consistent NULL handling:

  - Add NullBinary struct with Bytes []byte and Valid bool fields
  - Update scanNullBytes() to return NullBinary instead of []byte
  - Update ConvertValue for "varbinary" case to use NullBinary pattern
  
NULL values now properly return nil, empty values return []byte{}
 
Closes #166

### DIFF
--- a/trino/trino.go
+++ b/trino/trino.go
@@ -2533,7 +2533,11 @@ func (c *typeConverter) ConvertValue(v interface{}) (driver.Value, error) {
 		}
 		return vv.String, err
 	case "varbinary":
-		return scanNullBytes(v)
+		vv, err := scanNullBytes(v)
+		if !vv.Valid {
+			return nil, err
+		}
+		return vv.Bytes, err
 	case "tinyint", "smallint", "integer", "bigint":
 		vv, err := scanNullInt64(v)
 		if !vv.Valid {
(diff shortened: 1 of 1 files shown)


nineinchnick:
CLA submissions are processed manually by Martin every two or three weeks. The last batch was done about two weeks ago. You don't have to do anything, I keep track of this, and will merge this PR once the CLA is processed.

PR #168, opened 2025-11-15
Add ParseDSNToConfig function for DSN parsing support

### LINKED ISSUE #162: Add DSN parsing function for Trino URIs
Currently, the Go Trino client does not provide a built-in function to parse a Trino DSN/URI into a Config struct. Lack of a standard DSN parsing function means each client implementation may have to duplicate parsing logic. Could lead to inconsistencies and errors when parsing Trino URIs.

Flgado:
The Trino Go client currently only supports one-way DSN operations through Config.FormatDSN(), which converts Config structs to DSN strings. However, there was no way to parse existing DSN strings back into Config structs. This created several limitations:

- No roundtrip functionality: Users couldn't parse DSN strings they received from external sources
- Limited configuration flexibility: Applications couldn't dynamically parse and modify DSN configurations
- Missing tooling support: Development tools couldn't inspect or validate DSN configurations programmatically

This was an open gap in the client's functionality that prevented users from working with DSN strings in a bidirectional way.

Closes #162

### DIFF
--- a/trino/trino.go
+++ b/trino/trino.go
@@ -164,6 +164,8 @@ const (
 	defaultallowedOutOfOrder       = 10
 	defaultSpoolingDownloadWorkers = 5
 	defaulttrinoEncoding           = "json"
+	defaultSourceName              = "trino-go-client"
+	defaultKerberosServiceName     = "trino"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


trino/trino.go
```diff
@@ -209,7 +209,125 @@ type Config struct {
 	QueryTimeout               *time.Duration    // Configurable timeout for query (optional)
 }
 
-// FormatDSN returns a DSN string from the configuration.
+func ParseDSNToConfig(dsn string) (*Config, error) {
+	serverURL, err := url.Parse(dsn)
```
nineinchnick: Does this duplicate `newConn(dsn string)`? Maybe we can call it there and simplify setting up the connection by taking values from the Config struct.

trino/trino.go
```diff
@@ -429,60 +538,55 @@ func newConn(dsn string) (*Conn, error) {
...
-		parts := strings.SplitN(entry, mapKeySeparator, 2)
-		if len(parts) != 2 {
-			return nil, fmt.Errorf("trino: Malformed %s: %s", name, input)
-		}
-		key := parts[0]
-		value := parts[1]
-		if len(key) == 0 {
+func decodeMapHeader(name string, m map[string]string) ([]string, error) {
+	result := make([]string, 0, len(m))
+	for k, value := range m {
```
nineinchnick: Keep the key name
```suggestion
	for key, value := range m {
```

nineinchnick (approved):
This looks great, thank you!

nineinchnick:
Please squash commits and avoid unnecessary commit message prefixes.

Issue #176, opened 2026-04-14
Add support for Trino NUMBER type


Mandukhai-Alimaa:
The trino-go-client doesn't recognize the new NUMBER type introduced in Trino. When querying a NUMBER column, DatabaseTypeName reports VARCHAR instead of NUMBER.

https://trino.io/blog/2026/03/25/number-data-type.html

PR #177, opened 2026-04-17
Add support for Trino NUMBER type

Flgado:
# Summary
Adds support for Trino’s NUMBER type (Trino 480+) in the database/sql driver so queries no longer fail with an unsupported-type error and metadata reports the correct SQL type.

Close issue #176 

# Behavior
- `NUMBER` is handled like DECIMAL:` values are exposed as string / sql.NullString so precision is not reduced by float64.
- `ColumnTypeDatabaseTypeName()`  returns NUMBER (distinct from DECIMAL and VARCHAR).
- `ConvertValue` and `getScanType` include number alongside the existing string-based types; nested array(number) paths stay aligned with other string-like element types.

# Tests
TestTypeConversion: long fractional literal, long integer string, NaN, Infinity, -Infinity, plus nil and bogus inputs.

# Docs
README: DECIMAL and NUMBER documented together as string types with sql.NullString for nullable columns.

 #Notes
- Does not parse NUMBER into big.Float / big.Int; callers parse the string if they need arbitrary-precision math (same pattern as DECIMAL).


slachiewicz:
Sorry for the noise from my side — I filed a batch of issues before checking open PRs, and this one already covers #176.

I could not find a change to `X-Trino-Client-Capabilities` in the diff — it is still the hardcoded `"PARAMETRIC_DATETIME"` at `trino.go:1201`. From `ClientCapabilities.java`:

> Whether client supports the `NUMBER` type. When this capability is not set, the server returns `varchar` for `NUMBER` columns.

If that is right, a real server never sends a `NUMBER` column to this driver, so the new `getScanType` and `ConvertValue` branches would not be reached in practice and `ColumnTypeDatabaseTypeName` would keep reporting `VARCHAR` — the symptom #176 describes.

The unit tests would still pass, because they exercise `ConvertValue` with synthetic input rather than a real column, and the only integration-test change here is the LocalStack tag pin. So a query against an actual `NUMBER` column is the check that would settle it either way.

If it does need the capability, it is presumably one more value in that header, plus an integration test that selects a `NUMBER` column on Trino 480+ and asserts `DatabaseTypeName`. Happy to be wrong here — I have not run this branch, and you know the type path better than I do.

Separately: this pins LocalStack to `4.14.0`, and #182 pins it to `4.14` for the same reason. Whichever merges second will want to drop that hunk.

Flgado:
@slachiewicz Thank you! You’re right the changes have been applied. Please review them when you have a chance, @nineinchnick.

PR #180, opened 2026-07-30
Update to LZ4 v4

akgerber:
LZ4 has been updated to a new major version that maintains an identical interface for the function called in this repo:
https://github.com/pierrec/lz4

Also run `go mod tidy`.


nineinchnick:
Done in https://github.com/trinodb/trino-go-client/commit/f5a59664eca7aecb0852daa5051962fe9c60024a

PR #182, opened 2026-08-10
Upgrade ory/dockertest to v4 and pin LocalStack to 4.14

slachiewicz:
Two commits.

**Pin LocalStack to 4.14.** `localstack/localstack:latest` moved to the 2026.x line, which exits at startup unless `LOCALSTACK_AUTH_TOKEN` is set. Master's integration tests cannot start at all today because of this. The pin does not depend on the upgrade and can be cherry-picked on its own.

**Upgrade ory/dockertest to v4.** v4 drops the vendored Docker client, so the test plumbing moves onto the moby client types, and the pool takes a context and functional options.

What changes if you run the integration tests:

- Containers and the network left over from an earlier run are deleted at startup, and the fixtures are built fresh. v3 adopted and restarted them instead, whatever state or image tag they were in. v4 only reuses containers inside a single process, so there was no direct replacement for that.
- `-no_cleanup` still leaves the containers running after a run so you can inspect them. It no longer decides what the next run picks up.
- A setup that fails part way now removes the containers it had already started, instead of leaving them running and holding their ports.
- go 1.25.5 is required, so the CI matrix has a single Go entry.


.github/workflows/ci.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go: ['>=1.25','1.24.7']
+        go: ['>=1.25','1.25.5']
```
nineinchnick: We can drop 1.24.7 and I don't think we need to test multiple minor versions of the same major one.

trino/integration_test.go
```diff
@@ -204,75 +221,178 @@ func TestMain(m *testing.M) {
...
+	resource, err := pool.Run(ctx, repository, opts...)
+	if err != nil {
+		log.Fatalf("Could not start %s container: %s", repository, err)
+	}
+	containerTeardown = append(containerTeardown, resource.Close)
+
+	return resource
+}
+
+// adoptContainer takes over a container left behind by an earlier run started
```
nineinchnick: I'm not sure if we should be doing this. Containers left over without `-no_cleanup` sound like a fatal error and might not be healthy. We should require the user to manually clean up the resources, accepting the cost of setting up test fixtures from scratch.

nineinchnick (commented):
Remove the `Generated-by` advertisement from commit messages.

trino/integration_test.go
slachiewicz: Worth flagging that this predates the upgrade rather than arriving with it. v3's `pool.ContainerByName` filters the daemon's container list with `All: true`, and master calls it unconditionally — not gated on `-no_cleanup` — for both `trino-go-client-tests` and `localstack`, starting the container again if it was stopped. So master already adopts a container left by a crashed run, or by a run with a different `-trino_image_tag`. v4 has no equivalent, because its reuse registry is in-process only, which is why the same behaviour shows up here as hand-written code instead of one call.

Happy to drop it. There is one consequence to decide on first: setup errors go through `log.Fatalf`, which skips the teardown at the end of `TestMain`, so a health-check timeout or a failed `generateCerts` leaves the containers running. That is invisible today because the next run adopts them — without reuse it becomes a manual `docker rm` after every flaky setup. CI is unaffected either way, since each job gets a fresh runner.

Three ways to go:

1. Drop reuse and leave the leak — the next run stops and names the `docker rm` to run.
2. Same, plus tear down whatever had started when setup fails, with `-no_cleanup` still keeping a broken container for inspection.
3. Remove a leftover automatically and carry on, so fixtures are always rebuilt without any manual step.

I have 2 working locally, as a commit that removes reuse ahead of the v4 upgrade plus one for the teardown. Which would you prefer before I push?

nineinchnick: We can have both 2 and 3. We can start with 3

slachiewicz: Both, with 3 first — `232f378` removes leftovers by name and rebuilds from scratch, `098c302` tears down what a failed setup had started. `-no_cleanup` still keeps containers after a run for inspection; it just no longer decides what the next run reuses.

Checked locally: with both container names planted stale, `go test -race ./trino/` is green after logging the removal of each; an injected failure just after LocalStack starts leaves no container and no network, and the same failure under `-no_cleanup` keeps them for the next run to clear. Ctrl-C and `go test -timeout` still leak, since neither runs any teardown.

slachiewicz:
Looks like with moby client update to latest version we can simplyfy some parts also. I've signed up CLA and mail sent.

nineinchnick:
Thank you, I'll keep an eye on the CLA so you don't have to do anything once this PR gets approved.

trino/integration_test.go
```diff
@@ -262,21 +258,10 @@ func setupFatal(ctx context.Context, format string, v ...any) {
 	os.Exit(1)
 }
 
-// cleanupDocker removes the containers this run started, then the network they
-// were attached to, and finally closes the pool's Docker client.
+// cleanupDocker removes everything the pool is tracking. Close takes the
+// containers before the networks, which Docker refuses to remove while
+// containers are still attached, and closes the Docker client last.
 func cleanupDocker(ctx context.Context) error {
```
nineinchnick: Can we inline this now?

nineinchnick (approved):

PR #208, opened 2026-08-12
Error when a DSN password is used over HTTP

### LINKED ISSUE #205: DSN password is silently discarded over http instead of erroring
A password in the DSN is silently discarded when the scheme is `http`. From `newConn` ([`trino.go:588`](https://github.com/trinodb/trino-go-client/blob/master/trino/trino.go#L588)):

```go
var user string
if serverURL.User != nil {
    user = serverURL.User.Username()
    pass, _ := serverURL.User.Password()
    if pass != "" && serverURL.Scheme == "https" {
        c.auth = serverURL.User
    }
}
```

Refusing to send a password in cleartext is right. Doing it silently is not: the connection is still opened, the user name is still sent as `X-Trino-User`, and the query is attempted unauthenticated. The caller sees an unexplained 401, or worse, a session that succeeds as the wrong principal on a cluster that does not require authentication.

The Java client refuses explicitly instead. `TrinoUri.java`:

> TLS/SSL is required for authentication with username and password

## Suggested fix

Return an error from `ParseDSN` or `newConn` when a password is supplied with an `http` URL, with a message naming the cause. Anyone relying on the current behaviour is relying on their credentials being dropped, which is not a behaviour worth preserving compatibly.

Flgado:
close #205 

Reject username/password in an `http` DSN instead of dropping the password silently. TLS is required, same as the Java client.

Ps: Integration tests will fail until we will merge one of the PR's with the pinned localstack

### DIFF
--- a/trino/trino.go
+++ b/trino/trino.go
@@ -228,6 +228,9 @@ func ParseDSN(dsn string) (*Config, error) {
 	if err != nil {
 		return nil, fmt.Errorf("invalid DSN: %w", err)
 	}
+	if err := requireTLSForPassword(serverURL); err != nil {
+		return nil, err
+	}
 
 	query := serverURL.Query()
 	config := &Config{}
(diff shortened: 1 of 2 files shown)


slachiewicz:
Master now carries the LocalStack pin from #182, so the integration failure this PR mentions is fixed there. A rebase on master should turn CI green.

*This comment was created with AI assistance.*

nineinchnick (approved):

PR #209, opened 2026-08-12
Reject the accessToken named argument when forwarding is disabled

### LINKED ISSUE #204: accessToken named argument is inlined into query text unless ForwardAuthorizationHeader is set
Passing an access token as a named argument inlines it into the query text unless `ForwardAuthorizationHeader` is set, so the token is sent as SQL and recorded in Trino's query history.

## Reproduction

With default configuration:

```go
db.Query("SELECT ?", sql.Named("accessToken", "super-secret-jwt-value"))
```

What the server receives:

```
Authorization header sent: ""
query text sent to server: EXECUTE _trino_go USING 'super-secret-jwt-value'
```

## Cause

The interception is gated on the config flag ([`trino.go:1215`](https://github.com/trinodb/trino-go-client/blob/master/trino/trino.go#L1215)):

```go
if st.conn.forwardAuthorizationHeader && arg.Name == accessTokenConfig {
    token := arg.Value.(string)
    hs.Add(authorizationHeader, getAuthorization(token))
    continue
}
```

Without the flag there is no `continue`, so the argument falls through to the ordinary parameter path and is serialised as a string literal.

## Why it matters

Query text is durable in a way headers are not: it appears in the coordinator's query history, in event listeners, and in the Web UI. A credential that reaches it should be assumed disclosed, and there is no error or warning to indicate it happened.

The trigger is a single missing config flag, and the failure is invisible from the client — the query succeeds, and the parameter is simply the wrong thing.

More generally, any named argument the driver does not recognise falls through to the parameter path. That is reasonable for ordinary values, but the magic-name arguments are the ones most likely to carry secrets.

## Suggested fix

Recognise `accessToken` unconditionally, and either apply it as a header or reject it with a clear error when `ForwardAuthorizationHeader` is not set. Silently treating a credential as a query parameter is the one outcome that should not be possible.

Flgado:
Fixes #204

The `accessToken` named argument was only recognized when `ForwardAuthorizationHeader` was set. Without it, the token fell through to the normal parameter path and was inlined into the query text (`EXECUTE _trino_go USING '<token>'`), where Trino persists it in the query history, event listeners and Web UI — silently, with no error.

`accessToken` is now always recognized, and returns an error when forwarding is not enabled, so the token can never reach the query text. Also replaced the unchecked type assertion on the token value, which panicked on a non-string.

Note: this is a behavior change for anyone binding a query parameter literally named `accessToken`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,10 @@ This driver supports forwarding authorization headers by adding a [NamedArg](htt
 
 When enabled, this configuration will override the `AccessToken` set in the `Config` struct.
 
+Using the `accessToken` named argument without enabling `ForwardAuthorizationHeader` returns an
+error, so the token is never sent as part of the query text, where Trino would persist it in the
+query history.
+
 
 #### System access control and per-query user information
 
--- a/trino/trino.go
+++ b/trino/trino.go
@@ -109,6 +109,10 @@ var (
 
 	// ErrInvalidProgressCallbackHeader indicates that server did not get valid headers for progress callback
 	ErrInvalidProgressCallbackHeader = errors.New("trino: both " + trinoProgressCallbackParam + " and " + trinoProgressCallbackPeriodParam + " must be set when using progress callback")
+
+	// ErrForwardAuthorizationHeaderNotEnabled indicates that the accessToken named argument was used
+	// without enabling forwarding, which would otherwise send the token as part of the query text.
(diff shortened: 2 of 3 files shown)


slachiewicz:
Master now carries the LocalStack pin from #182, so the integration failure this PR mentions is fixed there. A rebase on master should turn CI green.

*This comment was created with AI assistance.*

nineinchnick (approved):

PR #215, opened 2026-09-14
Drop aws-sdk-go v1 from the integration tests

### LINKED ISSUE #188: Drop the end-of-support aws-sdk-go v1 dependency, used for two helper calls
`github.com/aws/aws-sdk-go` (v1) is a direct requirement in `go.mod`, but it is used for exactly two helper calls, both in `trino/integration_test.go`:

```go
UsePathStyle: *aws.Bool(true),
Bucket:       aws.String(bucketName),
```

Everything else on the S3 path already uses v2 — `aws-sdk-go-v2/config`, `aws-sdk-go-v2/credentials`, `aws-sdk-go-v2/service/s3`. The v1 import is the only thing keeping the module in the graph.

## Why it is worth removing

v1 is end-of-support. From its own README:

> :no_entry_sign: This SDK has reached end-of-support
>
> Per that announcement, as of 7/31/2025, the SDK has entered end-of-support. It will not receive further updates.

So it will not receive fixes, including security ones, and dependency scanners flag it. It is also a 322 MB module in the cache, carried for two function calls.

## The change

`aws-sdk-go-v2/aws` has the same helpers, so the import swap is direct:

```go
UsePathStyle: true,                  // *aws.Bool(true) allocates a *bool and immediately dereferences it
Bucket:       aws.String(bucketName),
```

Then `go mod tidy` drops the v1 requirement.

Test-only, so it does not affect what consumers link, but it does clear the module from `go.mod`, `go.sum` and any scan of this repo.

Happy to send a PR — it needs one integration run to confirm the spooling bucket is still created. I would keep it separate from #182 since it is unrelated to the dockertest upgrade.

slachiewicz:
The v1 SDK reached end of support on 2025-07-31 and was only used for `aws.Bool` and `aws.String` in the LocalStack bucket helper. The v2 `aws` package, already in the module graph, provides the same, so v1 drops out of `go.mod` for every consumer.

Fixes #188


Flgado (approved):

nineinchnick:
Done in 9b9bfc4e38302067467a9b4d94f973011b70fe45

PR #216, opened 2026-09-14
Move the integration tests into their own module

### LINKED ISSUE #210: Move the integration tests into their own Go module so consumers stop inheriting dockertest, moby and aws-sdk-go
`trino/integration_test.go` is in the same module as the driver, so every consumer of `github.com/trinodb/trino-go-client` carries its test-only requirements in their own module graph. Go tracks requirements per module, not per package or per build; there is no test scope.

Concrete cost, measured from [warpstreamlabs/bento](https://github.com/warpstreamlabs/bento), which uses the driver only through `database/sql`:

- At `v0.313.0`, `go mod graph` shows `github.com/trinodb/trino-go-client@v0.313.0 github.com/ory/dockertest/v3@v3.10.0`, and through it `docker/cli v25.0.3+incompatible`, a vendored 2019 Docker client. After bento migrated its own tests off dockertest v3, trino-go-client was the sole remaining path (warpstreamlabs/bento#965).
- At master (`f81af824`, after [PR #182](https://github.com/trinodb/trino-go-client/pull/182)), the edges are `ory/dockertest/v4@v4.0.0`, `moby/moby/api@v1.55.0`, `moby/moby/client@v0.5.1`, `moby/docker-image-spec@v1.3.1` and `aws/aws-sdk-go@v1.55.8` (end-of-support, [#188](https://github.com/trinodb/trino-go-client/issues/188)). Better versions, same class of problem.
- The `go 1.25.5` directive that dockertest v4 forced on this module is now the floor for every consumer, although the driver itself needs none of 1.25.

[#79](https://github.com/trinodb/trino-go-client/issues/79) explored this in 2023 and tried a separate *package*, which cannot help for the reason above. A separate *module* does: a nested `go.mod` (for example `trino/integration/go.mod` or `internal/itest/go.mod`) that requires the root module makes dockertest, moby, LocalStack and aws-sdk-go requirements of the test module only. Consumers of the root module never see them, and the root `go` directive can drop back to what the driver needs. The CI workflow would run the nested module explicitly (`go test` in that directory, or a `go.work`), since `go test ./...` at the root no longer reaches it.

I can open a PR if the shape is acceptable.

*This issue was created with AI assistance.*

slachiewicz:
Go tracks requirements per module, so every consumer of the driver inherited dockertest and the moby client through the integration tests in package `trino`. They now live in `integration/`, a separate module that requires the driver through a `replace` directive, together with the container config under `etc/` and the benchmarks. The root module keeps only what the driver imports; `go test ./...` at the root runs the unit tests without Docker or `-short`, and CI runs the integration module with `go test -C integration`. The unit job gains the minimum Go version so the driver still builds on it.

The tests import the driver as an external package, so the few unexported constants they used become the literal strings a user would write (`sql.Named("encoding", …)`, `X-Trino-Client-Tags`, `_trino_go`).

The module location is open to preference; `trino/integration` or `internal/itest` work the same way.

Verified: `go list -m all` at the root goes from 73 modules to 30, none of them dockertest or moby; `go test -race ./...` at the root and `go test -C integration -race ./...` against Trino latest and 372 pass.

Fixes #210


nineinchnick:
Can you rebase? In 7b387e2673ff6e738ec45f60e454452055d5a379 I bumped the minimum Go version to 1.25.5 because `github.com/klauspost/compress` requires it.

nineinchnick:
@slachiewicz I just did a major refactor of all tests. I recommend re-doing this from scratch, but it should be much easier now, as I fixed the overlap between unit and integration tests and separated them in the CI.

nineinchnick (approved):

Flgado (approved):

nineinchnick (approved):

Issue #181, opened 2026-08-03
Transaction statements (START TRANSACTION/COMMIT/ROLLBACK) and Begin() don't actually track Trino's transaction ID


Mandukhai-Alimaa:
`Conn.Begin()` returns `ErrOperationNotSupported`, so `database/sql`'s `Tx` interface is explicitly unsupported. But raw SQL transaction statements sent via plain `Exec`/`Query` have the same gap silently: `stmtResponse`/`queryResponse` have no field for Trino's `startedTransactionId`/`clearTransactionId`, so it's dropped on unmarshal, and nothing ever sends `X-Trino-Transaction-Id` back. So `START TRANSACTION` + DML + `COMMIT` as separate `Exec` calls run as unlinked, independently-autocommitted statements.

The client already does this for session properties (capturing`X-Trino-Set-Session`/`Clear-Session` and re-attaching `X-Trino-Session`, see `Conn.roundTrip` ~[trino.go:814-823](https://github.com/trinodb/trino-go-client/blob/master/trino/trino.go#L814)). Transactions would need the same: cache the transaction ID on `Conn`, echo it via `X-Trino-Transaction-Id`, and implement `Begin()`/`ConnBeginTx` on top of that instead of erroring.

PR #217, opened 2026-09-16
Add support for transactions

Flgado:
### How it works on the wire

Trino does not keep transaction state on a socket — every statement is its own HTTP request, and the transaction is identified by the `X-Trino-Transaction-Id` header. The server also refuses `START TRANSACTION` unless that header is present, which is how it detects whether a client understands transactions at all. So `BeginTx` sends the header as `NONE` to open one, `roundTrip` picks up the `X-Trino-Started-Transaction-Id` the server allocates, and every later request on that connection replays it. `X-Trino-Clear-Transaction-Id` clears it again.
`Conn` now implements `driver.ConnBeginTx` and `driver.SessionResetter`. The latter matters because a connection going back to the pool must not carry a finished transaction ID, or it would fail every later query against a transaction that no longer exists.

### Why this has to go through `BeginTx`, and why raw SQL is still rejected
Transaction control statements sent straight to the pool — `db.Exec("START TRANSACTION")` — are still rejected by the server, deliberately.
A `*sql.DB` is a connection pool. Each `db.Exec` borrows whatever connection is free, runs one statement, and hands it back. Since the transaction ID lives in a per-connection header bag, three raw `db.Exec` calls can easily land on three different connections: the first opens a transaction, the second runs in autocommit because it never sends the ID, and the third tries to commit a transaction its connection knows nothing about.
That is not theoretical. Driving the raw protocol by hand, sending one `INSERT` with the transaction ID and one without it — exactly what two pooled connections would do — then rolling back:
```
1) START TRANSACTION                -> txn 5ccadcda-...
2) INSERT 1 WITH the txn id         -> ok
3) INSERT 2 WITHOUT it              -> ok
4) ROLLBACK                         -> ok
5) count rows -> 1
```
One row survives a rollback that should have discarded everything, with no error anywhere. That silent loss of atomicity is the concern raised in #181, and the server's up-front rejection is the safer behavior, so this PR keeps it.
`sql.Tx` is the one thing that gives the needed guarantee: `database/sql` pins a single connection for the lifetime of a `Tx` (`Tx.dc` is owned exclusively until `Commit` or `Rollback`, and every statement gets it via `Tx.grabConn`). That is why support is exposed through `BeginTx` rather than by making raw `START TRANSACTION` work.
### Rollback after a failed statement
Trino aborts a transaction itself as soon as a statement inside it fails, and then refuses `ROLLBACK` with `TRANSACTION_ALREADY_ABORTED` for sessions carrying catalog-scoped state such as a role. `Rollback` tolerates that error, since the transaction is discarded either way, which is all the caller asked for. `Commit` still reports it, so a failed write is never mistaken for a successful one.


nineinchnick:
@Flgado I significantly refactored all tests. Please rebase, and there are two new tests verifying the _lack_ of transaction support, which you'll need to remove.

nineinchnick (approved):

Issue #187, opened 2026-08-11
Spooled segment acknowledgment errors are silently dropped


slachiewicz:
When the spooling protocol is in use, each downloaded segment is acknowledged with a `GET` to its `ackUri`. That acknowledgment runs in a goroutine whose errors are dropped ([`trino.go:1534`](https://github.com/trinodb/trino-go-client/blob/master/trino/trino.go#L1534)):

```go
//acknowledge the segment read
go func() {
    // TODO: handle ack erros
    ackReq, err := http.NewRequestWithContext(sf.ctx, "GET", sf.spooledMetadata.ackUri, nil)
    if err != nil {
        return
    }
    ...
```

A failed acknowledgment is invisible: the request error is discarded, and so is a non-2xx response.

## Why it matters

An unacknowledged segment stays in object storage until the spooling TTL expires it. On large result sets that is a real storage bill, and it is silent — nothing in the client or the query result indicates that segments were left behind. It also only shows up under exactly the conditions where it is hardest to notice: transient failures during big queries.

## Shape of the fix

- Retry a failed ack under the same policy as the main request path, which is why this is worth doing after the retry work is settled.
- Surface persistent failure somewhere observable rather than dropping it.
- A failed ack must **not** fail the query. The rows were already read correctly; the only loss is storage cleanup.

## Question

What is the right way to surface it? Options I can see:

1. A logger or callback on `Config`, alongside the existing `ProgressUpdater` pattern.
2. Reuse the existing warning path, if surfacing client-side warnings that way is acceptable.
3. Just count them and expose the count.

Option 1 fits the existing style best, but it adds public API for a fairly narrow condition, so I would rather ask than guess.

Part of #184.

<sub>This work was created with AI assistance.</sub>

nineinchnick:
Logs are annoying to handle in the driver. I'd prefer `ProgressUpdater`, so option 1.

Issue #33, opened 2022-01-04
Add tests for transactions


hashhar:
See https://github.com/trinodb/trino-go-client/issues/32#issuecomment-1004694778

This will probably require a connector that supports transactions if we want to test this end-to-end.

nineinchnick:
I wonder if anything from this PR would be useful: https://github.com/prestodb/presto-go-client/pull/66

PR #219, opened 2026-09-19
Rework the test suite

### LINKED ISSUE #34: Add tests for UPDATE and DELETE

nineinchnick:
The test suite had grown by accretion: server-backed tests lived in the unit file, the same fake spooling server was copy-pasted ten times, two assertion styles coexisted, and several tests waited on wall-clock sleeps or on the query deadline instead of on events. The unit suite took 26 seconds and failed under -short; the Docker suite spent 19 seconds streaming five million rows to prove segment ordering. Three tests could not fail, and TestMain mutated package-level timeouts and the default HTTP transport for every test in the process.

This is a series of 20 commits. The first six are refactors that change no assertions (each keeps the same test list and outcomes): move the server-backed tests out of the unit file, split both files by topic, serve fake responses from one shared coordinator, use testify with (expected, actual) order, lay out table tests the same way everywhere, and name the segment retry delay so tests can shorten it. The remaining commits change what is tested: -short means "no server", redundant and assertion-free tests are dropped or merged, unit tests wait for events instead of the clock, no state is shared between tests so they run in parallel, every assertion can fail, integration tests prove the same things with less server time, the container lifecycle detects the server version and cleans up on interrupt, and coverage is added for the driver branches nothing reached.

Four driver bugs surfaced while writing the tests, each fixed in its own commit with a reproducing test:

- A statement retried after a 502, 503 or 504 was sent without its body whenever the coordinator or a load balancer closed the connection, so the retry never reached the server.
- Spooled segment acknowledgements were dropped when the query finished before the acknowledging goroutine ran.
- The segment dispatcher reported an error and carried on with the bad segment, and an inline segment whose data was not a string crashed the process.
- The encoding, spooling_worker_count and max_out_of_order_segments named arguments crashed the process when given a non-string value.

Fixes #34 with UPDATE and DELETE tests on an Iceberg catalog, since the memory catalog cannot update rows. Addresses #33 by pinning down the current transaction contract (Begin fails locally, the server rejects transaction statements, no transaction header is sent), written so that a BeginTx implementation such as #217 extends the table. The integration files stay under trino/ with an integration_ prefix so #216 can move them into their own module with a rename.

Measured on the CI commands, from a clean Docker state:

| Run | Before | After |
|---|---|---|
| go test -short -race (no Docker) | 25.8 s, failing | 2.6 s, passing |
| Docker suite, Trino latest | 65 s | 36 s |
| Docker suite, Trino 372 | 27 s | 27 s |

CI now runs the unit suite as its own Docker-free job next to the integration matrix. CONTRIBUTING.md documents the flags, including running against an already started server with -trino_server_dsn.

Defects noticed but left out of scope, to be filed separately: Conn has no driver.Pinger so db.Ping never reaches the server, and acknowledgement errors are still ignored (#187).


nineinchnick:
@Flgado @slachiewicz FYI. Remember to pull the latest master before starting any new PRs :-)

Flgado:
Amazing refactor @nineinchnick ! well done 😲  !

PR #221, opened 2026-09-19
Send X-Trino-Time-Zone and read naive temporal values in that zone

### LINKED ISSUE #183: Client never sends X-Trino-Time-Zone, so naive timestamps can be silently wrong
The client never sends `X-Trino-Time-Zone`, but it parses timestamps without a time zone using the client machine's zone. The two ends therefore disagree about what zone a naive timestamp is in.

## What the protocol says

> `X-Trino-Time-Zone` — The timezone for query processing. Defaults to the timezone of the Trino cluster, and not the timezone of the client.

— [Trino client REST API](https://trino.io/docs/current/develop/client-protocol.html)

So with the header absent, the server evaluates `current_timestamp`, `now()`, `date_trunc` and similar in the **cluster's** zone.

## What the client does with the result

[`trino.go:3190`](https://github.com/trinodb/trino-go-client/blob/master/trino/trino.go#L3190) parses `timestamp`/`time` without zone as:

```go
t, err = time.ParseInLocation(layout, v, time.Local)
```

`time.Local` is the **client machine's** zone. Nothing reconciles the two.

## Effect

On a cluster running UTC, queried from a machine in `Europe/Warsaw`:

```sql
SELECT current_timestamp(6)
```

The server returns the wall-clock reading in UTC, and the client hands back a `time.Time` carrying the same wall-clock digits labelled `+02:00`. Converting it, comparing it to `time.Now()`, or writing it back all shift it by the offset. Nothing errors — the value is just wrong, and it is only wrong when the client and cluster zones differ, so it tends to survive local testing and appear in production.

The Java client sends the JVM default zone in this header, so both ends agree on the interpretation.

## Suggested fix

1. Send `X-Trino-Time-Zone` on every request, defaulting to `time.Local`'s IANA name.
2. Add a DSN parameter and `Config` field to override it, for callers who want to pin the cluster's zone or UTC.

Point 2 matters for anyone depending on today's behaviour: making the client suddenly send its own zone changes results for existing users, so an explicit setting gives them a way back.

## Before I start

Happy to implement this, but I would rather agree the shape first, since it is a behaviour change:

- Should the default be the client's zone (matching the Java client), or should it stay as-is and only apply when explicitly configured?
- Preferred name for the setting — `time_zone`?

nineinchnick:
The client never sent `X-Trino-Time-Zone`, so the server evaluated `current_timestamp`, `date_trunc` and every `TIMESTAMP` without a zone in the cluster's zone, while the driver labelled the returned wall-clock digits with `time.Local`. Whenever the two zones differed the `time.Time` was a different instant than the server produced, with no error to show for it, and only in deployments where client and cluster disagree, so the problem tended to surface in production rather than in local testing.

The header now goes out on every request, defaulting to the client's local zone like the JDBC driver and the CLI do, and the same zone is used to read `DATE`, `TIME` and `TIMESTAMP` values. A new `timezone` DSN parameter and `Config.TimeZone` field pin the zone for callers who want results independent of where the client runs. Since Go names the local zone "Local" unless `TZ` is set, the default id is recovered from `TZ` or the `/etc/localtime` symlink, falling back to the current offset, which Trino also accepts.

The per-statement `X-Trino-Time-Zone` named argument, which already reached the server, now also drives parsing, and a `SET TIME ZONE` reported through the `time_zone_id` session property is followed too, so the two ends stay in agreement after either kind of override.

Answers to the questions raised in the issue: the default is the client's zone, matching the Java clients, and the parameter is called `timezone` after the JDBC URL property and the CLI flag.

The first commit is a preparatory refactor with no behaviour change: it makes the parse location an explicit argument and adds an exported `Location` field to the `NullSliceTime` scanners, which callers invoke themselves and so cannot see the connection's zone.

Fixes #183

🤖 Generated with [Claude Code](https://claude.com/claude-code)


slachiewicz (approved):

trino/trino.go
```diff
@@ -1449,6 +1500,13 @@ func (st *driverStmt) exec(ctx context.Context, args []driver.NamedValue) (*stmt
 					st.user = headerValue
 				}
 
+				if arg.Name == trinoTimeZoneHeader {
```
Flgado, line 1503: Quick question on timezone after `SET TIME ZONE` as I am confuse..

If someone does `SET TIME ZONE 'Asia/Tokyo'` and then runs a query with `sql.Named("X-Trino-Time-Zone", "Europe/Paris")`, which timezone should we use when reading `DATE` / `TIME` / `TIMESTAMP` values?

Right now the named argument wins. I expected the session timezone from `SET TIME ZONE` to win, like on the server. Did I get that wrong?

Same question for `time_zone_id` in `session_properties` on the DSN: we send it, but we only start using it for parsing after a `SET TIME ZONE` comes back.
Should that connect-time property already apply?

nineinchnick: Executing the `SET TIME ZONE` and passing named arguments are equivalent operations. You'd use named arguments from a program using the driver, and users without direct access to the driver could issue the query. Trino servers are stateless, and the client keeps the state. That's why when you issue a `SET TIME ZONE` or other similar query, you just receive back the state you're supposed to keep and keep sending to the server.

Issue #199, opened 2026-08-11
Replace the 21 hand-written Null* types with generic equivalents


slachiewicz:
The driver exports 21 hand-written `Null*` struct types — a cross-product of element type and nesting depth: `NullSliceBool`, `NullSlice2Bool`, `NullSlice3Bool`, `NullSliceMap`, `NullSlice2Map`, and so on. Each has a near-identical `Scan`.

The matrix has a hard ceiling. `getScanType` stops at three dimensions, and the code says so ([`trino.go:2604`](https://github.com/trinodb/trino-go-client/blob/master/trino/trino.go#L2604)):

```go
// if this is a 4 or more dimensional array, scan type will be an empty interface
```

So a four-dimensional array is unscannable except as `interface{}`, and any element type outside the fixed set is unreachable no matter the depth.

Generics would cover the whole space: `NullSlice[T]`, `NullMap[K, V]`, and a `NullRow[T]` for ROW columns scanned into a struct — which is the same problem as #194 and probably wants designing with it.

**This should be purely additive.** Every existing type stays exported and working; the old ones can get deprecation comments, but removing them is a breaking change and should be out of scope.

<sub>This work was created with AI assistance.</sub>

Issue #194, opened 2026-08-11
ROW columns scan as bare interface{} with no field names


slachiewicz:
`getScanType` ([`trino.go:2531`](https://github.com/trinodb/trino-go-client/blob/master/trino/trino.go#L2531)) switches on the raw type name and has cases for boolean, the varchar family, varbinary, the integer family, real/double, the date/time family, `map` and `array` — but no `row`.

So a ROW column falls through to the `v == nil` branch and scans as a bare `interface{}`. `typeConverter.ConvertValue` ([`trino.go:2663`](https://github.com/trinodb/trino-go-client/blob/master/trino/trino.go#L2663)) only validates that the value is a slice and passes it straight through, so the caller gets positional values with no field names, and has to know the ROW's shape by index.

The Java client has `Row` and `RowField` carrying the field names, which are present in the `typeSignature` the driver already parses.

The useful version of this is probably scanning a ROW into a struct by field name, which overlaps with the generic null scanners in #199 — they may want designing together.

<sub>This work was created with AI assistance.</sub>

slachiewicz:
Before writing this, one decision is yours: carrying field names means the value a ROW column scans into can no longer be `[]interface{}`.

The shape I would propose is a `Row` value with `Names []string` and `Values []interface{}`, produced by a converter that recurses into the fields, so a VARBINARY or TIMESTAMP inside a ROW is decoded like a top-level column instead of reaching the caller as its JSON string (which `TestComplexTypes` in the integration module currently pins as the expected behaviour). Callers that type-assert `[]interface{}` today would break; the module is v0, so that is allowed with a changelog entry, but it is a break.

The alternative that keeps `[]interface{}` is a `NullRow[T]` scanner in #199 that maps struct fields positionally and never sees the names.

Which do you want?

nineinchnick:
`Row{Names, Values}` sounds better. We can make a breaking change here to reach feature parity with JDBC. I'd bump the version to v1 once we're done.

PR #223, opened 2026-09-20
Decode values of unknown types the way the Java client does

### LINKED ISSUE #145: Align UDT handling in trino-go-client with Java client’s Base64-encoding default
## Summary

Today the two official Trino clients diverge in how they handle User-Defined Types (UDTs):

* **trino-java-client**
  Defaults to serializing UDT values via a Base64 encoder, allowing arbitrary binary payloads to round-trip without error.
* **trino-go-client**
  Treats any UDT as an unsupported type and immediately raises an exception when encountering one.

This mismatch causes cross-language interoperability issues and surprises Go users who expect the same “just works” behavior as in Java. We should align the Go client’s behavior to match the Java client: default to Base64-encoding UDT values, and only error if encoding itself fails.

---

### trino-go-client

* **Unsupported by default**
  In `(*stmt).NextColumnType`, encountering a column with `type.kind == “user_defined”` will cause `NextColumnType` (or downstream type mapping) to return an error:

  ```go
  return nil, fmt.Errorf("unsupported column type: %s", tc.String())
  ```

### trino-java-client

* **Base64 encoder default**
  In `io.trino.client.StatementClient`, UDT payloads are represented as opaque `byte[]` values, which the client maps to Base64 strings in JSON. When rows are fetched, those JSON Base64 values are decoded into Java’s `byte[]` and can be consumed via `ResultSet.getBytes()`.
* **No error path**
  Unless the Base64 itself is malformed, the Java client will never raise a “unsupported type” error for UDT columns.

---

## Proposal

1. **Add a Base64 default converter for UDT columns in trino-go-client**

   * In the column-type detection logic (e.g. in `ColumnConverterForType`), detect `type.kind == user_defined` and return a converter that:

     1. Reads the raw payload bytes
     2. Calls `encoding/base64.StdEncoding.EncodeToString([]bytePayload)`
     3. Returns the resulting `string` to the caller

slachiewicz:
A column whose raw type is outside the driver's fixed list failed the whole query with `type not supported: "HyperLogLog"`, so `approx_set`, `qdigest`, `tdigest` and connector-defined types were unusable from Go. The Java client never had that gap: `JsonDecodingUtils` decodes any type it does not know as a base64 payload, treats `Color` as a string, and passes `KdbTree`, `BingTile` and `variant` through as JSON.

This mirrors that dispatch for top-level columns. Unknown raw types now decode their base64 payload to `[]byte` the same way varbinary does; `color` scans as a string; `KdbTree`, `BingTile` and `variant` reach the caller as the decoded JSON value. `Geometry` and `SphericalGeography`, already strings in `ConvertValue`, now also report `sql.NullString` as their scan type instead of `interface{}`. The raw type name Trino reports for `color()` is lowercase `color`, unlike the mixed-case geospatial types, which is why the string case uses that spelling.

Element types nested inside arrays and maps are left as they were: they already pass through undecoded, and changing them is a separate decision.

The issue title talks about user-defined types; Trino has none, and what actually fails is any type name the driver does not list. Nested-type decoding and a named `Row` value are tracked in #194 and #199.

Verified: `go test -C integration -race ./... -trino_image_tag=latest` and `-trino_image_tag=372` → `ok`; `approx_set`, `color` and `bing_tile` exist on both images.

Fixes #145


nineinchnick:
Whoops, I hadn't noticed this one and merged #228. I've compared it with this PR, and there are minor differences we'll address later on. Like we don't yet advertise `variant` support, and the unit tests here are better with fake-coordinator.

